PR #3, opened 2017-09-19
CBF File Functionality Tests

cydanil:
- Update the dependencies needed for the creating CBF files;
- Create tests for that functionality.

### DIFF
--- a/bin/euxfel_h5tool.py
+++ b/bin/euxfel_h5tool.py
@@ -50,6 +50,7 @@ def check_filenames_are_files(filenames):
         assert os.path.isfile(filename)
     return filenames
 
+
 def convert_path_to_list_of_files(path):
     # convert path into list of filenames where required
 
@@ -64,8 +65,6 @@ def convert_path_to_list_of_files(path):
         filenames = arguments['PATH']
 
 
-
-
 if __name__ == "__main__":
 
     arguments = docopt.docopt(__doc__, version=version)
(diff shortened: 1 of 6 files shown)


bin/euxfel_h5tool.py
```diff
@@ -76,25 +75,21 @@ def convert_path_to_list_of_files(path):
     # Iterate over files for options that process single files
 
     # structure
-    if arguments['structure'] == True:
+    if arguments['structure'] is True:
```
jwiggins: `if arguments['structure']`... It's PEP8 plus all kinds of other conventions

euxfel_h5tools/__init__.py
```diff
@@ -1 +1 @@
-from .lib import stat, rec_print_h5_level, h5_to_cbf
+from .lib import (h5_to_cbf, rec_print_h5_level, stat, np_to_cbf)
```
jwiggins: Parentheses not needed

euxfel_h5tools/lib.py
```diff
@@ -114,6 +113,14 @@ def rec_print_h5_level(ds, indent=0, maxlen=100):
         else:
             print(" "*indent + k)
 
+
+def np_to_cbf(np_array, index=0):
+    """Given a 3D np array, convert it to a CBF data object"""
+    img_reduced = np_array[index, ...]
+    cbf_out = fabio.cbfimage.cbfimage(header=header, data=img_reduced)
```
jwiggins: `header` is not defined here.

euxfel_h5tools/lib.py
```diff
@@ -122,13 +129,16 @@ def h5_to_cbf(in_h5file, cbf_filename, index, header=None):
         header = {}
     try:
         tmpf = h5py.File(in_h5file, 'r')
+        paths = list(tmpf["METADATA/dataSourceId"])
+        image_path = [p for p in paths if p.endswith(b"image")][0]
+        images = tmpf[image_path + b"/data"][index]
```
jwiggins: Why the `b''` here? Maybe wrap the `p`s above with `str()`?

euxfel_h5tools/test_lib.py
```diff
@@ -1,7 +1,7 @@
 import os
+import pytest
```
jwiggins: `pytest ` is not part of the standard library, so it should not be here with the stdlib imports

PR #18, opened 2018-03-26
Allow H5File to act as a context manager itself

takluyver:
This allows the `H5File` class to be used as a context manager, so you can write:

```python
with H5File(...) as sensor_data:
   ...
```

i.e. the separate `open_H5File` function is no longer needed. I've left that function there in case other code is already using it, but if you know that nothing else does use it yet, we can get rid of it.

### DIFF
--- a/euxfel_h5tools/reader.py
+++ b/euxfel_h5tools/reader.py
@@ -205,10 +205,19 @@ class H5File:
     def close(self):
         self.file.close()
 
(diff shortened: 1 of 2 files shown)


tmichela:
Great,
Nothing uses `open_H5File` in the package at least, and probably nothing outside (in theory the normal entry point should be the RunHandler class).
I would get rid of it.

takluyver:
Done :-)

PR #19, opened 2018-04-03
Generating mock data for tests

takluyver:
This code produces an HDF5 file with the same structure (as far as I can tell) as one of the example files from a single AGIPD module. The main data is left empty, so the file is < 1MB. Then it uses this to replicate one of the tests that currently needs a real data file.

Future work, when I have access to more data:

- Check if this data layout needs any changes for the latest files
- Add code for generating example non-detector data files
- Add tests for whole runs of mock data


tmichela:
One important layout details, is that datasets are written per chunks. I think that for detector data is it by default 16, but could be in theory of any size.
so this means that if you have 250 entries in your file, you would have 6 extra values on each datasets filled with zeros.

takluyver:
I haven't tried to replicate the chunking (I've left it up to h5py's defaults), but I have replicated the datasets that have a dimension of 256 with the last 6 left unfilled. That's what this code does (for one such field):

```python
pc = f.create_dataset('INSTRUMENT/SPB_DET_AGIPD1M-1/DET/7CH0:xtdf/header/pulseCount',
                    (256,), 'u8', maxshape=(None,))
pc[:250] = 64
```

tmichela:
ah right, thanks!

tmichela:
Great, thanks! You can merge when you think it's ready

PR #24, opened 2018-04-27
Test Python 3.4-3.6 on Travis

takluyver:
This would have caught my mistake that led to #23 .

As a side effect, tests also run faster, so long as Travis has enough free workers to run them concurrently.

I don't think we gain anything from defining a docker container with pip and pytest for this. Travis already has those things. I'd like to get rid of the Dockerfile from this repo.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,11 @@
-sudo: required
+language: python
+python:
+  - 3.6
+  - 3.5
+  - 3.4
 
-services:
-  - docker
-
-before_install:
-  - make build-docker-image
+install:
+  - make install
 
 script:
-  - make ci-docker
+  - make test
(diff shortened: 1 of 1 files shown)


fangohr:
I'd like to keep the Dockerfile: what I really like is that one has (i) a fully defined environment and (ii) exactly the same environment that is used by the testing system. This will also be useful when we get third party contributions that fail on Travis but didn't fail for the contributor - we can then ask them to make things pass in the container (even if they don't have a linux box).

The docker build time should be small (I am hoping for clever caching from Travis). 

[This may be of particular value when working on OSX and Windows.]

Happy to discuss - maybe I am missing something in the discussion above.

takluyver:
We agreed to do the tests natively in Travis, but leave the docker file in place. That's what this PR does - I suggested removing the Dockerfile but didn't do it yet. So I'll merge this.

PR #26, opened 2018-04-30
Pandas integration

takluyver:
This adds two new methods, with the same parameters for files and runs.

Get a series (column) by device name and data key:

```python
run.get_series('SA1_XTD2_XGM/DOOCS/MAIN', 'beamPosition.iyPos.value')
```

Get a dataframe by a glob of device+key names:

```python
run.get_data(fields=['*_XGM/*.i[xy]Pos'])
# With the timestamps:
run.get_data(fields=['*_XGM/*.i[xy]Pos*'], timestamps=True)
```


takluyver:
Suggestion from @tmichela : keep device and key names separate when specifying columns for a dataframe, like this:

```python
run.get_data(fields=[('*_XGM/*', '*.i[xy]Pos')])
```

A bit more verbose, but consistent with how they're used elsewhere.

takluyver:
@tmichela are you happy for this to be merged as well? There's doubtless more that could be done, but I haven't been doing anything with it for a couple of weeks, and it's probably more interesting now to try to get people using it than to guess what else could be useful.

PR #27, opened 2018-05-09
Geometry (LPD)

takluyver:
This adds code for reading geometry information, and assembling detector images based according to the geometry, based on the code in femDataSplitter to do something similar with a live stream of data.

This is currently only for LPD, but it should be possible to extend the same mechanism to handle other detectors, and if necessary other geometry file formats.

It also includes an 'approximate' geometry so you can get a rough look at your data even if you don't have the proper geometry information. We have a few pieces of code that do similar approximate assemblies; the advantage of this one is that you can use the approximate geometry in the same way as real geometry data.

To help me figure out some issues, I also wrote a method to show the geometry on a plot without any data:

lpd_geom


tmichela:
Is this ready to be merged?

takluyver:
I think it is, yes.

@cydanil this is the PR for using hdf5 geometry files; have a look at the example notebooks.

PR #28, opened 2018-05-28
Sample geometry & nbval

takluyver:
Automatically check that the notebooks still run without error (skipping the one that requires real data on Maxwell).


tmichela:
Why is the approximate geometry removed? Maybe it was said this morning, sorry if I missed it...

takluyver:
It came up in discussion with Steffen on Friday that the quadrants are more mobile than I thought - the central hole can be opened in either direction (shifting the modules clockwise or anticlockwise). So the approximate geometry is less useful, and Steffen is keen to get people using real geometry.

I got the detector group's permission to share the March 18 geometry data instead, so this repository has all the necessary information to try out some real geometry. I'm still not sure whether to add that as a data file within the package (loaded by something like `LPDGeometry.march2018()`) or just as an example in the docs, as I've currently done.

PR #34, opened 2018-05-30
Add methods to get labelled array objects

takluyver:
While working on an example use case Hans sent me, I found that accessing the data train-by-train was very slow. So I added some methods to get a field at a time, as a labelled array. This is similar to the methods I already added to get a pandas series, but it handles multidimensional data using [xarray](https://xarray.pydata.org/en/stable/).

* The advantage of the labelled array over plain numpy arrays is that it makes it easy to align data by train IDs: `xarray.align(arr1, arr2, join='inner')` will return arrays with only the trains that are in both. It also lets you do things like `arr.mean('train')` to average along the train axis.
* Labelled arrays are, naturally, based on numpy arrays, so you can do anything you would do with a numpy array.
* This loads all the requested data into memory. On Maxwell, that should be OK for most data except big detector data.
* Like the other methods, `.get_array()` exists both on file objects and on a run directory. The run directory method concatenates arrays from several sequence files.
* Future extension: retrieve a [DataSet](http://xarray.pydata.org/en/stable/data-structures.html#dataset), xarray's multidimensional equivalent to a pandas DataFrame.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -384,6 +384,10 @@ class H5File:
         key: str
             Key of parameter within that device, e.g. "beamPosition.iyPos.value"
             or "header.linkId". The data must be 1D in the file.
+        extra_dims: list of str
+            Name extra dimensions in the array. The first dimension is
+            automatically called 'train'. The default for extra dimensions
+            is dim_0, dim_1, ...
         """
         name = self._make_field_name(device, key)
 
@@ -659,6 +663,10 @@ class RunDirectory:
         key: str
             Key of parameter within that device, e.g. "beamPosition.iyPos.value"
             or "header.linkId". The data must be 1D in the file.
+        extra_dims: list of str
+            Name extra dimensions in the array. The first dimension is
+            automatically called 'train'. The default for extra dimensions
+            is dim_0, dim_1, ...
         """
         seq_arrays = [f.get_array(device, key, extra_dims=extra_dims)
                       for f in self.files
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -675,7 +675,7 @@ class RunDirectory:
                       for f in self.files
                       if device in (f.control_sources | f.instrument_sources)]
 
-        return xr.concat(sorted(seq_arrays, key=lambda a: a.coords['train'][0]),
+        return xr.concat(sorted(seq_arrays, key=lambda a: a.coords['trainId'][0]),
                          dim='trainId')
 
     def _assemble_sequences(self):
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -375,6 +375,9 @@ class H5File:
     def get_array(self, device, key, extra_dims=None):
         """Return a labelled array for a particular data field.
 
+        The first axis of the returned data will be the train IDs.
+        Datasets which are per-pulse in the first dimension are not supported.
+
         Parameters
         ----------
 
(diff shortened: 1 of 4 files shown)


tmichela:
is the train-by-train very slow also when you filter the sources? I also experience slowness, but more on the long run: first train are coming "fast" then it is getting slower as we process large number of train.
Also experienced that on Maxwell the data reading was slower (maybe it has longer delay on accessing disk). We should maybe have an other -more optimized- was of loading the data on memory designed for Maxwell.

tmichela:
Sorry this is not really related to the PR, should I add an issue for this?

tmichela:
does it makes sens to add a test for `get_array`?

karabo_data/reader.py
```diff
@@ -371,6 +372,58 @@ def append_ctrl_data(key, value):
...
+                pulse_id = pulse_id[index_ds[:] != 0]
+                index = pd.MultiIndex.from_arrays([index, pulse_id],
+                                                  names=['trainId', 'pulseId'])
+        else:
+            raise ValueError("Unknown data source %r" % data_src)
+
+        if extra_dims is None:
+            extra_dims = ['dim_%d' % i for i in range(data.ndim - 1)]
+        dims = ['train'] + extra_dims
+        return xr.DataArray(data, dims=dims, coords={'train': index})
```
tmichela: maybe `trainId` or `train_id` instead of `train`?

takluyver: There's a minor wrinkle that some instrument data is recorded more than once per train. In that case (if the train ids are not unique), it builds a hierarchical index with levels `trainId` and `pulseId`. So I called the dimension 'train' as a bit of a fudge around that possibility. Maybe we should think of a better name.

karabo_data/reader.py
```diff
@@ -371,6 +372,58 @@ def append_ctrl_data(key, value):
...
+            index_ds = self.index['trainId']
+            index = index_ds[index_ds[:] != 0]
+            data = ds[:len(index)]
+        elif data_src.startswith('INSTRUMENT'):
+            ix_path = "/{}/trainId".format(data_src)
+            index_ds = self.file[ix_path]
+            index = index_ds[index_ds[:] != 0]
+            data = ds[index_ds[:] != 0, ...]
+            if len(np.unique(index)) != len(index):
+                raise ValueError("Train IDs are not unique for %s" % data_src)
```
tmichela: The message confuses me. TrainIds are always unique, it is the property that can have several values for a given train ID. Maybe something like: "%s has more that 1 data point per trainId" % data_src ?

takluyver: Thanks, that's clearer.

karabo_data/reader.py
```diff
@@ -371,6 +372,58 @@ def append_ctrl_data(key, value):
...
+        data_path = "/{}/{}".format(data_src, key.replace('.', '/'))
+        ds = self.file[data_path]
+
+        # Get the index
+        if data_src.startswith('CONTROL'):
+            index_ds = self.index['trainId']
+            index = index_ds[index_ds[:] != 0]
+            data = ds[:len(index)]
+        elif data_src.startswith('INSTRUMENT'):
+            ix_path = "/{}/trainId".format(data_src)
```
tmichela: are we sure that all `INSTRUMENT` data has a `trainId` dataset? Why don't we rely on `INDEX` group as for the `CONTROL` data?

takluyver:
As suggested, I've made it get the train IDs for the array index from the `INDEX` section of the file. I've re-run the SASE3 example that motivated this, and the plots produced appear identical to before.

tmichela:
LGTM!

PR #35, opened 2018-05-31
Add karabo-bridge-serve-files command

takluyver:
Prompted by discussion with @cydanil . Make it easy to stream data from files using the karabo bridge interface.


tmichela:
LGTM

cydanil:
Excellent, thanks!

takluyver:
We'll still need to make some changes to conform to the newer version of the protocol we're figuring out, but that's a separate question from the command line to launch it.

PR #40, opened 2018-06-01
Add test for iterating over selected fields

takluyver:
I was experimenting with a way to make selecting keys from a device more efficient. Unfortunately, it didn't make much difference, but this test I added is useful anyway.


tmichela:
LGTM

PR #43, opened 2018-06-06
Smoketest hdf5_file_info()

takluyver:
This function is a [big red block](https://codecov.io/gh/European-XFEL/karabo_data/src/a60995c12996590233c913df5aa18b145510c518/karabo_data/utils.py#L118) in the coverage at the moment, and it's easy to add a basic test for it.

The other changes are moving some pytest fixtures so that they're accessible to tests outside the `test_reader_mockdata` module.

### DIFF
--- a/karabo_data/tests/conftest.py
+++ b/karabo_data/tests/conftest.py
@@ -1,0 +1,40 @@
+import os.path as osp
+import pytest
+from tempfile import TemporaryDirectory
+from . import make_examples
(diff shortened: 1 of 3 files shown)


takluyver:
The PR tests are now passing. I fixed another test in master, so the branch tests don't pass, but the merge is OK.

tmichela:
LGTM

PR #46, opened 2018-06-11
More options for the .trains() method

takluyver:
Following an email from Dmitry, here are some proposed additions to the `.trains()` method:

- Selecting sources and keys gains the ability to use glob syntax, as I used for making pandas dataframes. E.g. if you only want detector data, you could write:

```python
for tid, data in run.trains([('*/DET/*', 'image.data')])
```

- In combination with this, you can opt to skip any trains which don't have data for all the sources you requested:

```python
for tid, data in run.trains([('*/DET/*', 'image.data')], partial=False)
```

- You can also specify a range of train IDs:

```python
for tid, data in run.trains(train_range=range(10004, 100024))
```

The design of all this is up for discussion. Of the code Dmitry sent us, this is the bit I'm trying to make it easier to do:

```python
for trainID in trainIDs[takeTrainsBegin:takeTrainsEnd+1]:
#    print("Processing train",trainID)
    
    tid, train_data = run.train_from_id(trainID)
    NoData = False
    for dev in sorted(train_data.keys()):
#        print(dev, end='\t')
        try:
#            print(train_data[dev]['image.data'].shape)
            train_data[dev]['image.data']
        except KeyError:
            print("No image.data for trainID",trainID)
            NoData = True
            break
    if NoData:
        continue
```

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -62,6 +62,52 @@ class FilenameInfo:
             self.description = "Unknown data source ({})", datasrc
 
 
+def _normalize_data_selection(selection, dataset):
+    """Normalize selected data fields
+
+    We offer different ways to select sources and keys from the data.
+    This function normalises the different inputs to a set containing
+    (source, key) tuples.
+
+    dataset is meant to be a H5File or RunDirectory object
+    """
+    if isinstance(selection, set):
+        return selection
+
+    res = set()
+    if isinstance(selection, dict):
+        # {source: {key1, key2}}
+        # {source: {}} -> all keys for this source
+        for source, keys in selection.items():#
+            if source not in (dataset.control_sources | dataset.instrument_sources):
+                raise ValueError("Source {} not in this run".format(source))
+
+            for k in (keys or dataset._keys_for_source(source)):
+                res.add((source, k))
+
+    elif isinstance(selection, list):
+        # [('src_glob', 'key_glob'), ...]
+        for src_glob, key_glob in selection:
+            matched = set()
+            for source in (dataset.control_sources | dataset.instrument_sources):
+                if not fnmatchcase(source, src_glob):
+                    continue
+
+                for key in dataset._keys_for_source(source):
+                    if fnmatchcase(key, key_glob):
+                        matched.add((source, key))
+
+            if not matched:
+                raise ValueError("No matches for pattern {}"
+                                 .format((src_glob, key_glob)))
+
+            res.update(matched)
+    else:
+        TypeError("Unknown selection type: {}".format(type(selection)))
+
+    return res
+
+
 class H5File:
     """Access an HDF5 file generated at European XFEL.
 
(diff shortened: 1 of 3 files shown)


karabo_data/reader.py
```diff
@@ -129,70 +175,134 @@ def _parse_data_src(source):
...
-                table.visititems(append_data)
 
-            sec, frac = str(time()).split('.')
-            timestamp = {'tid': int(self.train_ids[train_index]),
-                         'sec': int(sec), 'frac': int(frac)}
-            data.update({'metadata': {'source': device, 'timestamp': timestamp}})
+            train_data[device]['metadata'] = {
+                'source': source,
+                'timestamp': ts,
+                'timestamp.tid': int(self.train_ids[train_index]),
```
tmichela: better `train_id` instead of `int(self.train_ids[train_index])` since it is used everywhere else. Also why the casting? Does it return non-`int` type? if it is the case then I would do it line `215 train_id = int(self.train_ids[train_index])`

takluyver:
A fancier option I came up with for selecting trains:

```python
run.trains(range=TrainID[10004:10034])  # By train ID
run.trains(range=TrainIndex[4:34])   # By index in the run (or file)
```

That would make code using it more concise, but I worry that it's a bit too 'magical'. The implementation would also require a metaclass, which may make it harder for potential contributors to understand.

tmichela:
Also other idea: Why not using the `range` syntax in `train()`

```python
for tid, data in run.trains(10, 100, 5):  # index
    process(data)

for tid, data in run.trains(10000323, 10000550, 10):  # trainId
    process(data)
```

We can either:
- use a flag to distinguish `indexes` from `ids`,
- or since the `index` should be in the range `[0, X000]` and the `trainIds > 999999` we could extrapolate from the input
- or have 2 similar function, one for index the other for id
?

takluyver:
I don't want to take up three positional arguments for options that I think won't be used that much - I imagine it will be more common to specify some sources/keys and ask for only trains with complete data.

They could be keyword arguments like `start=10004, stop=10024`.

Yet another possibility is to use a tuple, like `train_range=(10004, 10024)`. Or I could make simpler objects which allow `range=TrainID(10004, 10024)` rather than `range=TrainID[10004:10024]`.

> use a flag to distinguish indexes from ids,

Possible.

> or since the index should be in the range [0, X000] and the trainIds > 999999 we could extrapolate from the input

As you can probably guess, I don't like that. ;-)

> or have 2 similar function, one for index the other for id

Possible, but a bit inelegant.

tmichela:
I would be more in favor either of kwargs if you don't like positional arguments: `start`, `stop`, `step`, maybe something like `selector=trainId` `selector=index`.
or tuple: `train_range=(10004, 10024)`, `index_range=(10, 20, 2)`, `pulse_range(1000, 2000)` (?)

You were also asking for a replacement of `partial` to make it non-negative, maybe simply `partial_ok=True`
?

takluyver:
My last commit adds caching of the index arrays. I *think* this improves performance for iterating through trains.

It's hard to test performance of reading files, because caching makes the results super variable, but for iterating over all trains & all sources in an FXE dataset, the best time I have seen without this caching was 6:50, and the best time with caching was 5:30. The difference is well within the variability of each, but the measurements I've taken do look like there's an improvement on average.

The CPU time in user mode, which is much more consistent, also improves significantly, from about 200s to about 110s. Here it's much easier to say that the difference is real.

takluyver:
Selecting out the detector data (`('*/DET/*', 'image.data')`) for `.trains() gives almost an order of magnitude improvement in access time: from 5-15 minutes to ~1 minute, even though the detector images are the largest part of the data.

takluyver:
Updated syntax following discussion with @tmichela and Marijan:

```python
for tid, data in run.trains([('*/DET/*', 'image.data')], by_index[20:], require_all=True)
```

That means:

- Select all detector devices, and the `image.data` key from each
- Give me from the 20th train in the file to the end
  - There is also `by_id` to select trains by train ID
- Skip trains which don't have all the requested data

karabo_data/reader.py
```diff
@@ -61,6 +61,52 @@ def __init__(self, path):
         else:
             self.description = "Unknown data source ({})", datasrc
 
+class _SliceContstructor(type):
```
tmichela: typo "_SliceConstructor" ?

tmichela:
LGTM.
Maybe @fangohr wants to have a look before we merge.

takluyver:
Thanks, fixed the typo.

Hans, if you have a chance, the bit we're least certain of is the `by_index[20:]` and `by_id[:10024]` syntax for slicing trains. This offers a concise & unambiguous way to write a lot of different selections we can imagine, but using these special `by_index` and `by_id` objects may confuse people who know some Python.

We felt that the advantages outweighed the disadvantages, but it was a fairly close call - at least for me.

takluyver:
Also, if we want to make any changes to the idea of selecting data by glob patterns like `[('*/DET/*', 'image.data')]` , it's much easier to do that before people start using this PR.

fangohr:
Thanks Thomas and Thomas. Can we discuss in the Monday team meeting, please?

PR #47, opened 2018-06-11
WIP: attempt to align interface with karabo_bridge

tmichela:


karabo_data/export.py
```diff
@@ -124,26 +141,61 @@ def _serialize(self, data):
...
+
+        Parameters
+        ----------
+        data : dict
+            Contains train data. The dictionary has to follow the karabo_bridge
+            protocol structure:
+            - keys are source names
+            - values are dict, where the keys are the parameter names and
+              values must be python built-in types or numpy.ndarray.
+        metadata : dict, optional
```
takluyver: The file reading code puts metadata in the source data dictionary, like the bridge protocol used to do. There should probably be something somewhere in the code that pulls it out of there and sends it correctly.

karabo_data/export.py
```diff
@@ -124,26 +141,61 @@ def _serialize(self, data):
...
+              {
+                  'timestamp': 1234.567890  # float
+                  'timestamp.sec': '1234'  # str
+                  'timestamp.frac': '567890000000000000'  # str
+                  'timestamp.tid': 1234567890  # int
+              }
+
+            'timestamp' is the Unix epoch
+            'timestamp.sec' is the seconds of Unix epoch
+            'timestamp.frac' is the fractional part of Unix epoch with
```
takluyver: Wording nitpick: the 'Unix epoch' is the zero point of the measurement system. The measurements can be called  'Unix time' or 'seconds since the Unix epoch'. https://en.wikipedia.org/wiki/Unix_time

karabo_data/export.py
tmichela: Right! I will update using that.
You won't like this :) but I was thinking if it would then make sens to align the karabo_data `RunDicrectory.train()` to the karabo_bridge, i.e. getting `data, metadata` instead of `trainId, data`...

takluyver: I wondered if you were going to say that, and you're right, I don't like it. ;-) The return type is a critical part of the API, and changing it unexpectedly would break people's code in confusing ways.

What we could do is add a parameter like `separate_metadata=True` to change it from `trainId, data` to `trainId, data, metadata`. Or make a `.trains2()` method, though that's somewhat ugly.

tmichela: I still believe no-one uses it :)
Right, I have to think about it but I still think it makes sens to have an interface matching the `karabo_bridge`'s output. `bridgely=True`

takluyver: Dmitry already showed us code using it, and we've shown it to quite a few people.

I think it probably makes sense to converge the interfaces via optional parameters, but maybe converge towards `train_id, data, metadata` rather than just `data, metadata`? As we've seen in writing the glimpse and monitor tools, it's not entirely convenient to pull it out of metadata every time.

takluyver: I had a chat with Hans, and he agreed with you that it's still OK to make breaking changes, though he suggested having a parameter so people can go back to the old format rather than changing all their code.

I still think we should try to converge karabo_data and karabo_bridge to provide `train_id, data, metadata`. Although once we've got 3-tuples, I start to wonder if we should be returning objects instead. Let's chat about this at some point.

tmichela: I was not really arguing, I just have 2 concerns,
1. Having an identical interface as the `karabo_bridge` can be useful to just plug a script that has been used online to process data. The `ZMQStreamer` class is supposed to do that but I wonder if it also makes sens to be able to not use this extra layer.
2. There is still this `metadata` key that is added to the data. Should we check if the data already contain that key? print a warning, raise?... I guess we can ignore that for now...
I don't have strong opinion here, just thinking out loud.

fangohr: Let's talk - maybe over lunch?

PR #48, opened 2018-06-12
Add an example of extracting xarray arrays

takluyver:
This is documented as part of the API, but I didn't add an example in the demo notebook before.


takluyver:
Merging this because it only changes docs, and I want to use it in my talk this afternoon.

takluyver:
Wrong button...

Issue #31, opened 2018-05-28
Check that train IDs are contiguous


takluyver:
Another request from the meeting this morning. Some data has had problems with missing trains - maybe especially where sequences files break?

Was this trains entirely missing from the files' index? Or trains which are in the index but have no associated data from a given source?

Other notes:

- train_ids monotonically increasing by one: True, False
- exclude DAQ dummy data: single image per train, trainID==0
  [train ID is zero until data is received]
- trainID should increment by one, throughout all sequence files until
  detector stops recording [new dummy data], or run is stopped.
  [there were some skipped trainIDs between sequences]

takluyver:
Another sanity check we could do is to check the correspondence between the indexes for instrument data and the `trainId` datasets inside the instrument data. They appear to be two redundant ways of specifying the same information.

fangohr:
Good idea.

I think it would be good to have a special 'verify' or 'sanity' mode that can check the consistency of the files in multiple ways. ITDM has interest in this and would likely extend such a mode with additional checks. We also know that some users have already implemented similar work. 

I would imagine that we have the functionality and a command line tool or option which works through the file and carries out all the consistency checks we can think of, and then reports which (if any) of those have failed and which have passed. 

It's okay if this is slow.

takluyver:
Another integrity check we can do: we have found at least some corrected data files with the index pointing to data beyond the end of the relevant dataset.

E.g. the detector image data in `/gpfs/exfel/exp/FXE/201701/p002026/proc/r0078/CORR-R0078-LPD01-S00000.h5` .

PR #54, opened 2018-06-18
Start of a framework for checking data files

takluyver:
See #31. I'm not totally happy with the design of this, but it's a starting point to experiment with.

It can be used at the command line like this:

```
$ python3 -m karabo_data.validation /gpfs/exfel/exp/FXE/201701/p002026/proc/r0078
Checking run directory: /gpfs/exfel/exp/FXE/201701/p002026/proc/r0078
Validation failed!

Index referring to data (9300) outside dataset (2480)
  dataset: INSTRUMENT/FXE_DET_LPD1M-1/DET/3CH0:xtdf/image/length
  file: /gpfs/exfel/exp/FXE/201701/p002026/proc/r0078/CORR-R0078-LPD03-S00003.h5

Index referring to data (9300) outside dataset (2480)
  dataset: INSTRUMENT/FXE_DET_LPD1M-1/DET/3CH0:xtdf/image/trainId
  file: /gpfs/exfel/exp/FXE/201701/p002026/proc/r0078/CORR-R0078-LPD03-S00003.h5
...
```

### DIFF
--- a/karabo_data/validation.py
+++ b/karabo_data/validation.py
@@ -1,3 +1,4 @@
+from functools import partial
 import numpy as np
 import os
 import sys
@@ -30,7 +31,35 @@ class FileValidator:
     def run_checks(self):
         self.problems = []
         self.check_indices()
+        self.check_trainids()
         return self.problems
+
+    def record(self, msg, **kwargs):
+        self.problems.append(dict(
+            msg=msg, file=self.filename, **kwargs
+        ))
+
+    def check_trainids(self):
+        ds_path = 'INDEX/trainId'
+        train_ids = self.file.file[ds_path][:]
+
+        if (train_ids == 0).any():
+            first0 = train_ids.tolist().index(0)
+            if not (train_ids[first0:] == 0).all():
+                self.record(
+                    'Zeroes in trainId index before last train ID',
+                    dataset=ds_path,
+                )
+            nonzero_tids = train_ids[train_ids != 0]
+        else:
+            nonzero_tids = train_ids
+
+        if len(nonzero_tids) > 1:
+            if not (nonzero_tids[1:] > nonzero_tids[:-1]).all():
+                self.record(
+                    'Train IDs are not strictly increasing',
+                    dataset=ds_path,
+                )
 
     def check_indices(self):
         for src in self.file.instrument_sources:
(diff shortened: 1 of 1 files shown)


karabo_data/validation.py
```diff
@@ -30,8 +31,36 @@ def validate(self):
...
+        self.check_trainids()
         return self.problems
 
+    def record(self, msg, **kwargs):
+        self.problems.append(dict(
+            msg=msg, file=self.filename, **kwargs
+        ))
+
+    def check_trainids(self):
+        ds_path = 'INDEX/trainId'
```
tmichela: now that I think about it, I believe that `INDEX/trainId` should not contain zeroes at all, the ones that can contain zeros in datasets are in `INSTRUMENTS/...` but they don't always contain this field. Here we can maybe simply check if there are zeroes at all.

karabo_data/validation.py
```diff
@@ -30,8 +31,36 @@ def validate(self):
...
+                self.record(
+                    'Zeroes in trainId index before last train ID',
+                    dataset=ds_path,
+                )
+            nonzero_tids = train_ids[train_ids != 0]
+        else:
+            nonzero_tids = train_ids
+
+        if len(nonzero_tids) > 1:
+            if not (nonzero_tids[1:] > nonzero_tids[:-1]).all():
```
tmichela: I would assume that they need to increase monotonically (by 1) but we better check with IT.

karabo_data/validation.py
takluyver: At least some of the older datasets have the train IDs written in chunks, so there are zeroes at the end. E.g. `/gpfs/exfel/exp/SPB/201730/p900008/raw/r0273/RAW-R0273-DA01-S00000.h5`. It seems like the newer files don't do this, but it's not clear when that changed or if it's a feature we can rely on. `karabo_data` is also built to handle trailing 0s there.

tmichela:
is this feature ready?

takluyver:
I think it's a reasonable framework to start with. We can add more checks to it as we go.

I had a vague idea for a system where each possible problem would be a separate class, with an idenfitier you could pass to disable it along with any sub-checks. But I couldn't work out how it would all fit together, and I suspect it may be more complexity than we need. So I think the framework that we have in this PR is a reasonable starting point.

takluyver:
Thanks!

Issue #52, opened 2018-06-14
Better error messages


takluyver:
Henry just reminded me that the error messages when you pass an incorrect source or key are very cryptic - we spit out internal errors from HDF5 with no information about how to correct it. We should improve on this.

PR #55, opened 2018-06-20
Better error messages on incorrect source or property name

takluyver:
Previously, mistyped names got errors like:

```
KeyError: 'Unable to open object (component not found)'
```

This PR makes error messages like:

```
SourceNameError: This run has no source named 'SA1_XTD2_XGM/DOOCSa/MAIN'.
See run.all_sources for available sources.


PropertyNameError: No property 'data.intensityTD' for source 'SA1_XTD2_XGM/DOOCS/MAIN'
```

Closes gh-52

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -24,7 +24,9 @@ import xarray as xr
 
 
 __all__ = ['H5File', 'RunDirectory', 'RunHandler', 'stack_data',
-           'stack_detector_data', 'by_id', 'by_index']
+           'stack_detector_data', 'by_id', 'by_index', 'SourceNameError',
+           'PropertyNameError',
+          ]
 
 
 RUN_DATA = 'RUN'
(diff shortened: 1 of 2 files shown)


fangohr:
That's great, thank you.

Here is suggestion: how about adding a link to a command line command to see the sources of a file/run, for example: replacing ``See run.all_sources for available sources.`` with something like ``Use "lsxfel --showsources" to list all sources (or check attribute "run.all_sources")`` ?

I imagine that people may see this error message when they have not just written the code (but maybe try to run an existing script on new data). At that point, they will not know at which point to modify the script to print the attribute, but running a shell command will be much easier.

takluyver:
That probably makes sense, though I'd leave the Python snippet as the primary suggestion. If you open a terminal, it's not guaranteed that `lsxfel` will be available on PATH, whereas your Python code has already got `karabo_data` loaded.

PR #56, opened 2018-07-06
Don't generate timestamps when reading data

takluyver:
From discussions a couple of weeks ago, the files don't currently store any timestamps, so we can't even give an approximate time for a given train. We hope that in the future, timestamps from the time server will be recorded in the file.

In the meantime, I think we shouldn't generate timestamps when reading the file. This makes the metadata different from that given by `karabo_bridge`: you can't assume that timestamps are present. I think it's better to have that difference than to fill in the gap with misleading values.

### DIFF
--- a/karabo_data/export.py
+++ b/karabo_data/export.py
@@ -14,7 +14,6 @@ from argparse import ArgumentParser
 import os.path as osp
 from queue import Queue
 from threading import Event, Thread
-from time import time
 
 import msgpack
 import numpy as np
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -19,7 +19,6 @@ import numpy as np
 import os.path as osp
 import pandas as pd
 import re
-from time import time
 import xarray as xr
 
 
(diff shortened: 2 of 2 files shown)


fangohr:
Looks good to me. @tmichela - any comments an the reasoning for removing the time stamp info here?

tmichela:
The dummy timestamp exists because some tools use it to be able to sort results. For example, Onda is using it and is a mandatory parameter in this case (it won't work if the timestamp is missing).

I would be in favor of keeping this until we have timestamps stored in files.

takluyver:
If there are specific tools like Onda that need timestamps but don't need them to be accurate, I think we should either fix those tools or push the workaround closer to them (`stream_data_with_dummy_timestamps_for_onda()` :wink: ). I think it's misleading to generate incorrect timestamps, i.e. they don't remotely represent when the data was collected. If we haven't got a timestamp, we should send the data without it and let tools decide what that means.

takluyver:
I've added a `dummy_timestamps=True` parameter to the streaming machinery so we can generate these timestamps before sending the data over ZMQ. But it has to be explicitly requested.

I deliberately haven't added it to the CLI at the moment; it's easy to do if necessary, but I want this to be a bit awkward, to make it really clear that these timestamps are not meaningful.

takluyver:
@tmichela how do you feel about the approach I described in the comment above?

takluyver:
Thanks :-)

PR #58, opened 2018-07-11
Add a function for streaming data from files.

zhujun98:
It would be nice to have a convenient function to stream data from files.

### DIFF
--- a/karabo_data/export.py
+++ b/karabo_data/export.py
@@ -206,7 +206,8 @@ def serve_files(path, port, **kwargs):
     streamer = ZMQStreamer(port, **kwargs)
     streamer.start()
     for tid, train_data in data.trains():
-        streamer.feed(train_data)
+        if train_data:
+            streamer.feed(train_data)
     streamer.stop()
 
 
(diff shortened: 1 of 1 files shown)


karabo_data/export.py
```diff
@@ -188,20 +188,33 @@ def feed(self, data, metadata=None):
...
-        data = H5File(args.path)
+        data = H5File(path)
 
-    streamer = ZMQStreamer(args.port)
+    streamer = ZMQStreamer(port, **kwargs)
     streamer.start()
     for tid, train_data in data.trains():
-        streamer.feed(train_data)
-
+        if train_data:
```
takluyver: Have you come across some case where this `if` isn't satisfied?

zhujun98: Yes, the train_data could be empty. I found that in the run that Dmitry used for test - r0078. The code crashed with the empty train_data:

  File "/home/zhujun/karabo_data/karabo_data/export.py", line 45, in run
    interface.send_multipart(self.buffer.get())
  File "/usr/local/lib/python3.6/dist-packages/zmq/sugar/socket.py", line 438, in send_multipart
    return self.send(msg_parts[-1], flags, copy=copy, track=track)
IndexError: list index out of range

PR #59, opened 2018-07-12
Add examples from XPD with slow data

takluyver:
As requested on Redmine: https://in.xfel.eu/redmine/issues/27290


fangohr:
Great, thank you. Did we have the XGM team looking over the notebook already? If they are happy with it, then I am happy with it.
(There is a balance between how much detail we want to provide on basic pandas operations - there isn't much in at the moment, but that will benefit more experienced users and keeps the document short; so overall this is probably the best approach.)

takluyver:
Theo agreed that we could share the XGM example from run 150 with only the per-train data - this is based on an example notebook I prepared for them a few weeks ago. Joakim has looked over the second part with the PES example.

takluyver:
I'm taking that as a "LGTM" ;-)

PR #66, opened 2018-08-10
Timing scripts & consolidate code for globbing sources and keys

takluyver:
The changes to matching patterns for dataframe columns make it use the same machinery as the train-oriented methods. This leads to a minor speedup, which I expect to be more pronounced for files with large numbers of datasets, because it can avoid walking the whole structure of the hdf5 file.


takluyver:
OKed offline by Thomas M.

PR #67, opened 2018-08-10
Some more tests for the reader module, rewritten stack_detector_data

takluyver:
The rewritten `stack_detector_data` should, I hope, be clearer and more robust. Errors are raised rather than printed.

@tmichela is there a reason we need both `stack_data` and `stack_detector_data`? I can see that `stack_data` is potentially more flexible to things other than detector modules, but do we have a real use case besides detector modules?

I'd also like to try to design consistent APIs for filtering data by device names - the `only` and `xcept` parameters on these functions work in a different way to other filtering mechanisms we provide elsewhere.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -1153,6 +1153,15 @@ RunHandler = RunDirectory
 def stack_data(train, data, axis=-3, xcept=()):
     """Stack data from devices in a train.
 
+    For detector data, use stack_detector_data instead: it can handle missing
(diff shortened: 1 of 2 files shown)


tmichela:
great, thanks!

I don't remember exactly why both function exists (probably thought it might be usefull)...
I also agree with aligning filtering interface,
lgtm

takluyver:
Rewritten the stack_data function as well

PR #74, opened 2018-09-11
Writing selected data back to HDF5 file

takluyver:
This adds an interface to write selected data back to an HDF5 file in the same format (so it can be loaded with karabo_data again). Use it like this:

```python
from karabo_data import RunDirectory
run = RunDirectory('/gpfs/exfel/exp/SA1/201830/p900025/raw/r0150')
run.select('SA1_XTD2_XGM/*').write('/gpfs/exfel/exp/SA1/201830/p900025/scratch/test_xtd2_r150.h5')
```

This writes a single file - it doesn't try to create sequence files or split up different sources into different files.

The data is copied into the new file, it doesn't use references to the source files like the virtual dataset work. That means you can take the file outside of Maxwell and load it. But if you try to use this to make a file with all detector data, it will be a very big file. The virtual datasets are a better way to provide a combined view of detector data.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -899,6 +899,18 @@ class DataCollection:
         print('\tControls')
         [print('\t-', d) for d in sorted(ctrl)] or print('\t-')
 
+    def write(self, filename):
+        """Write the selected data to a new HDF5 file
+
+        You can choose a subset of the data using methods
+        like :meth:`select` and :meth:`select_trains`,
+        then use this write it to a new, smaller file.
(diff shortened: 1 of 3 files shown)


tmichela:
If you can add a test, that would be great :)

otherwise, LGTM

fangohr:
Agree that this looks good, and writing a single file may also be just the right thing - good to start with this in any case.

I agree some tests would be good, too.

takluyver:
Added one test

PR #78, opened 2018-09-21
Test ebad

ebadkamil:
1) metadata dictionary should look like?

{''*/DET/0CH*": {'source': xxx , 'timestamp.tid':xxx, 'timestamp.frac': xxx, 'timestamp.sec':xxx,   }}

2)
metadata[src]['timestamp.tid'] = int(metadata[src]['timestamp.tid'])

if this is not implemented then C++ client says "timestamp.tid" is of MAP type.

### DIFF
--- a/karabo_data/export.py
+++ b/karabo_data/export.py
@@ -26,224 +26,224 @@ __all__ = ['ZMQStreamer', 'serve_files']
 
 
 class REPInterface(Thread):
-	def __init__(self, context, port, buffer):
-		super(REPInterface, self).__init__()
-		self.context = context
-		self.port = port
-		self.buffer = buffer
-		self._stop_event = Event()
+    def __init__(self, context, port, buffer):
+        super(REPInterface, self).__init__()
+        self.context = context
+        self.port = port
+        self.buffer = buffer
+        self._stop_event = Event()
 
-	def run(self):
-		interface = self.context.socket(zmq.REP)
-		try:
-			interface.bind('tcp://*:{}'.format(self.port))
+    def run(self):
+        interface = self.context.socket(zmq.REP)
+        try:
+            interface.bind('tcp://*:{}'.format(self.port))
 
-			while not self.stopped():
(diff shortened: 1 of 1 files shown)


zhujun98:
```
metadata[src]['timestamp.tid'] = int(metadata[src]['timestamp.tid'])

if this is not implemented then C++ client says "timestamp.tid" is of MAP type.
```
I don't understand what do you mean byb MAP type

metadata
--------
source, , , string
timestamp, , , double
timestamp.frac, , , string
timestamp.sec, , , string
timestamp.tid, , , uint64_t

ebadkamil:
> ```
> metadata[src]['timestamp.tid'] = int(metadata[src]['timestamp.tid'])
> 
> if this is not implemented then C++ client says "timestamp.tid" is of MAP type.
> ```
> 
> I don't understand what do you mean by MAP type
> ## metadata
> 
> source, , , string
> timestamp, , , double
> timestamp.frac, , , string
> timestamp.sec, , , string
> timestamp.tid, , , uint64_t


accessing the containerType via 
for(auto &v : data.metadata)
{
auto container = v.second.containerType(); // returns "MAP" when key ( v.first) is  'timestamp.tid'
}

zhujun98:
```
accessing the containerType via 
for(auto &v : data.metadata)
{
auto container = v.second.containerType(); // returns "MAP" when key ( v.first) is  'timestamp.tid'
}
```
It returns an empty string because it is a scalar data. I just double checked it.

karabo_data/export.py
```diff
@@ -117,20 +117,32 @@ def stop(self):
             self._interface = None
 
     def _serialize(self, data, metadata=None):
+
+        for key,val in data.items():
+            if 'metadata' in val:
+                val['metadata']['timestamp.tid'] = int(val['metadata']['timestamp.tid'])
```
takluyver: This chunk can be dropped now - I fixed it in a different place in #79.

PR #81, opened 2018-09-28
AGIPD geometry

takluyver:
This is a first pass on the machinery for creating and applying geometry for AGIPD. See [the example notebook](http://nbviewer.jupyter.org/github/European-XFEL/karabo_data/blob/8709124daec8b00a187bacdd5906e497a8d8066a/docs/AGIPD%20Geometry.ipynb) to get an overview of assembling data with it.

Assembling the data is significantly harder because we're no longer assuming it's axis-aligned, so it involves a scipy interpolation function to map it onto the output array. This is quite slow. We may want a fast assembly option that assumes one input pixel maps to one output pixel.

I think the best option would be to do azimuthal integration on the panels separately and combine the results. This is most accurate, and avoids interpolating twice. It looks like PyFAI may have some support for this.

I haven't yet tried to integrate this with the hierarchical geometry descriptions we use for LPD.


karabo_data/geometry2.py
```diff
@@ -0,0 +1,326 @@
...
+    def centre(self):
+        return self.corner_pos + (.5 * self.ss_vec * self.ss_pixels) \
+                               + (.5 * self.fs_vec * self.fs_pixels)
+
+    def to_crystfel_geom(self, p, a):
+        name = 'p{}a{}'.format(p, a)
+        c = self.corner_pos
+        return CRYSTFEL_PANEL_TEMPLATE.format(
+            name=name, p=p,
+            min_ss=(a * 64), max_ss=((a * 64) + 63),
```
tmichela: maybe use `self.ss_pixels` here instead of numbers?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,326 @@
...
+            ss_vec=_crystfel_format_vec(self.ss_vec),
+            fs_vec=_crystfel_format_vec(self.fs_vec),
+            corner_x=c[0], corner_y=c[1], coffset=c[2],
+        )
+
+
+class AGIPD_1MGeometry:
+    pixel_size = 2e-7
```
tmichela: Can you add a comment with unit?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,326 @@
...
+    def __init__(self, modules):
+        self.modules = modules  # List of 16 lists of 8 fragments
+
+    @classmethod
+    def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
```
tmichela: How did you define `asic_gap` and `panel_gap`?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,326 @@
...
+
+    def position_all_modules(self, data):
+        """Assemble data from this detector according to where the pixels are.
+
+        Parameters
+        ----------
+
+        data : ndarray
+          The last three dimensions should be channelno, pixel_y, pixel_x
+          (lengths 16, 256, 256).
```
tmichela: `(16, 512, 128)`
aren't pixels x and y the other way around?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,326 @@
...
+        """Plot data from the detector using this geometry.
+
+        Returns a matplotlib figure.
+
+        Parameters
+        ----------
+
+        modules_data : ndarray
+          Should have exactly 3 dimensions: channelno, pixel_y, pixel_x
+          (lengths 16, 256, 256).
```
tmichela: `(16, 128, 512)`

karabo_data/geometry2.py
takluyver: People have told me that the pixels at the edges of the asics are twice as wide as the others. As there are two edges next to each other, I took that to mean there is effectively a 2 pixel gap between adjacent asics. This seems to work pretty well from my experiments with CrystFEL.

`panel_gap=29` is from trial and error working with CrystFEL.

karabo_data/geometry2.py
takluyver: x and y are confusing in terms of data; I've now switched these docstrings to CrystFELs fs/ss terminology.

takluyver:
I've implemented 'snapping' the geometry to a pixel grid, which speeds up assembling an image from ~7 seconds to ~0.1 seconds. I've updated the notebook to illustrate this:

http://nbviewer.jupyter.org/github/European-XFEL/karabo_data/blob/ade66133150c6fd4589e1dc691941117160de1b0/docs/AGIPD%20Geometry.ipynb

takluyver:
@tmichela if you try to load a geometry file that's missing modules/asics, it will currently throw a KeyError when it tries to look up the missing piece. We can always revisit that if we get a geometry file with missing bits.

fangohr:
@takluyver  - nice demo (http://nbviewer.jupyter.org/github/European-XFEL/karabo_data/blob/ade66133150c6fd4589e1dc691941117160de1b0/docs/AGIPD%20Geometry.ipynb). Can we include this in the repo somewhere as developer or more technical documentation?

takluyver:
Also, I just remembered that the AGIPD geometry has inherited the CrystFEL
coordinate scheme, which reverses the x and z axes relative to the XFEL
scheme we use for LPD. It shouldn't make a difference for azimuthal
integration, but it might be good to ensure that they use coordinates the
same way.

Furthermore, the new code adopts CrystFEL's convention of storing
coordinates in pixel units, which is convenient for working with data,
while the LPD code has coordinates in metres. We should at least ensure
that the units are clear.

tmichela:
1. After struggling a bit on giving the right quadrant origins, I think we should have a different origin reference than "the first pixel of the first ASIC of the first module". This gives a different origin for each quadrant.  we should think of a more generic origin that works the same for each quadrants.

2. would be nice to have a "perspective" option to allow giving the quadrant position with a view toward the beam (Crystfel standard) or toward the detector (EuXFEL standard)

takluyver:
Quadrant origins: I don't think there's any right answer here that will make it intuitive for everyone. Someone might want to measure the positions of the innermost corners next to the hole, or the outermost corners, or the bottom left of each, with left in either coordinate system. Then you need further information to say which way round the data for each module needs to fit in.

I think I recall that AGIPD can also be used in a 'windmill' layout, with two of the quadrants rotated 90 degrees. If so, that would introduce further complications.

I think CrystFEL has found a pragmatic solution: give the location of the first pixel in each quadrant and vectors describing where each pixel dimension goes from there. This allows a lot of flexibility in layout without having to `if/else` much code.

takluyver:
Obviously we can have further methods to convert any other useful description of the geometry into this scheme. But as the canonical representation in the code, I like the 'first pixel' scheme.

tmichela:
I also think that it makes a lot of sens, and I would not change the representation in the code.

But rather adding a user interface that hides this, because I've seen people struggling with this. You need expert knowledge of the detector to be able to place the origins (which is different for every quadrant)... Also I am not sure what interface will be better :) maybe instead we can give more detailed information about the detector layout in the documentation?

takluyver:
I don't see any way to really hide the need for expert knowledge here - I think documenting it clearly is the best we can do on that front.

takluyver:
I just started working on the difference in coordinate systems, pulled up the relevant docs to check, and I misremembered. Both CrystFEL and EuXFEL (from Steffen's technical note) use a right-handed coordinate system with the +z direction going along the beam away from the light source. So there's nothing to be done in terms of the geometry. If there's a need to display images flipped (as if you were looking from the sample to the detector), the assembled array can be reversed with `arr[:, ::-1]`.

takluyver:
I think this is in good shape if someone wants to review it.

takluyver:
Ping for review :-)

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        Returns (size_x, size_y), (centre_x, centre_y)
+        """
+        corners = []
+        for module in self.modules:
+            for tile in module:
+                corners.append(tile.corners())
+        corners = np.concatenate(corners)[:, :2]
+
+        # Find extremes, add 20 px margin
+        min_xy = corners.min(axis=0).astype(int) - 20
```
antarcticrainforest: For consistency, wouldn't it be better to also get rid of this margin?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        Parameters
+        ----------
+
+        modules_data : ndarray
+          Should have exactly 3 dimensions: channelno, pixel_ss, pixel_fs
+          (lengths 16, 512, 128). ss/fs are slow-scan and fast-scan.
+        """
+        from matplotlib.cm import viridis
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+        from matplotlib.figure import Figure
```
antarcticrainforest: Would it be better to use pyplot instead of the OO API. I guess people (including myself) are more familiar pyplot. So I'd vote for creating the figure objects with pyplot. Or is there a reason why this is a bad idea?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        less accurate, but can assemble data into a 2D array more efficiently,
+        because it doesn't do any interpolation.
+        """
+        new_modules = []
+        for module in self.modules:
+            new_tiles = [t.snap() for t in module]
+            new_modules.append(new_tiles)
+        return AGIPD_1M_SnappedGeometry(new_modules)
+
+class AGIPD_1M_SnappedGeometry:
```
antarcticrainforest: What about making this the default class? So that you don't have to call the snap method to access the snapped geometry?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+            out = np.nanmax(tmp, axis=0)
+
+        return out, centre
+
+    def _plotting_dimensions(self):
+        """Calculate appropriate dimensions for plotting assembled data
+
+        Returns (size_x, size_y), (centre_x, centre_y)
```
antarcticrainforest: I think this would be better if it would be (size_y, size_x), (centre_y, centre_x) to be consistent with the _plotting_dimension method in the snapped class. I've already had some confusion/issues with it

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+
+                affine_transform(tile_data, transform, offset=offset, cval=np.nan,
+                                 output_shape=size_yx, output=tmp[i * 8 + j])
+
+        # Silence warnings about nans - we expect gaps in the result
+        with warnings.catch_warnings():
+            warnings.simplefilter("ignore", category=RuntimeWarning)
+            out = np.nanmax(tmp, axis=0)
+
+        return out, centre
```
antarcticrainforest: So the centre that is returned here is fortran style (x,y)? but the data isn't?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+            for j, tile in enumerate(module):
+                tile_data = tiles_data[j]
+                # Offset by centre to make all coordinates positive
+                y, x = tile.corner_idx + centre
+                h, w = tile.pixel_dims
+                out[..., y:y+h, x:x+w] = tile.transform(tile_data)
+
+        return out, centre
+
+    def _plotting_dimensions(self):
```
antarcticrainforest: I know it might sound silly, and it doesn't really matter, but I think a method name of something like get_dimensions is better than plotting_dimensions, since it's not only about plotting anymore.

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        Parameters
+        ----------
+
+        modules_data : ndarray
+          Should have exactly 3 dimensions: channelno, pixel_ss, pixel_fs
+          (lengths 16, 512, 128). ss/fs are slow-scan and fast-scan.
+        """
+        from matplotlib.cm import viridis
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+        from matplotlib.figure import Figure
```
antarcticrainforest: Same as above, what about pyplot?

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        modules = []
+        for p in range(16):
+            tiles = []
+            modules.append(tiles)
+            for a in range(8):
+                d = geom_dict['panels']['p{}a{}'.format(p, a)]
+                tiles.append(AGIPDGeometryFragment.from_panel_dict(d))
+        return cls(modules)
+
+    def write_crystfel_geom(self, filename):
```
antarcticrainforest: I think here could be room for additional keyword arguments with information on things like clen, res, adu_per_eV or even more. Without those hdfsee won't work on the created geometry file.

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+        self.modules = modules  # List of 16 lists of 8 fragments
+
+    @classmethod
+    def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
+        """Generate an AGIPD-1M geometry from quadrant positions.
+
+        This produces an idealised geometry, assuming all modules are perfectly
+        flat, aligned and equally spaced within their quadrant.
+
+        The quadrant positions are given in pixel units, referring to the first
```
antarcticrainforest: Just out of curiosity how are the quadrant numbers defined? 1 is the upper left corner when looking into the beam? And then going counterclockwise? Which would mean that q1=mod 0-3; q2=mod 4-7; q3=mod 8-11; q4=mod 12-15?

tmichela: yes

karabo_data/geometry2.py
```diff
@@ -0,0 +1,483 @@
...
+                corner_x = quad_corner[0] + x_orient * (64 + asic_gap) * a
+                tiles.append(AGIPDGeometryFragment(
+                    corner_pos=np.array([corner_x, corner_y, 0.]),
+                    ss_vec=np.array([x_orient, 0, 0]),
+                    fs_vec=np.array([0, y_orient, 0]),
+                ))
+        return cls(modules)
+
+    @classmethod
+    def from_crystfel_geom(cls, filename):
```
tmichela: Can this class be extended to work with other geometry format or it does not really makes sens?

if so, it would maybe be nice to have something like:
```python
from_file(filename, format='crystfel')
```

takluyver: It might make sense to support other file formats; my plan was to have a separate method for each, like `from_crystfel_geom` and `from_euxfel_geom`. There would also be corresponding methods for writing geometry files, like `write_crystfel_geom`, `write_pyfai_distortion`...

I prefer this to a `format=` parameter, because it's easy to tab-complete methods in a notebook or an IDE, and because some formats might want parameters which may not make sense for all formats.

karabo_data/geometry2.py
takluyver: Good point, will do. I'll still leave a 1px margin here, to ensure that rounding differences don't try to access beyond the edge of the array.

karabo_data/geometry2.py
takluyver: Yeah, I'll see if I can switch everything to `y, x`. It's an unfortunate discrepancy between accessing the array by `row, column` and referring to plotting coordinates as `x, y`.

karabo_data/geometry2.py
antarcticrainforest: Yes that sounds like a good idea! Let's do that then

karabo_data/geometry2.py
antarcticrainforest: Ok

karabo_data/geometry2.py
takluyver: There certainly is room, but I suggest we work those out in a later pull request. For now, the file is written with a comment at the top explaining that it's not usable as is.

karabo_data/geometry2.py
takluyver: Yup, the quadrants have to be in the same order as the data to be assembled, and the data is stacked by module number. I'll see if I can clarify this in the docstring.

karabo_data/geometry2.py
antarcticrainforest: yes I'd prefer to have an additional axis keyword argument for sub-plots and other things

takluyver:
OK, I've reworked the API so there's only one class that's meant to be used directly (the 'snapped' geometry class still exists, but it's an implementation detail). The `_fast` methods use the 'snapped' geometry, while `position_modules_interpolate` assembles the data the slow way. I've got rid of `plot_data_interpolate` because it doesn't seem that useful to interpolate if you're just drawing a plot.

karabo_data/geometry2.py
```diff
@@ -258,68 +258,79 @@ def position_all_modules(self, data):
...
+        """
+        if self._snapped_cache is None:
+            new_modules = []
+            for module in self.modules:
+                new_tiles = [t.snap() for t in module]
+                new_modules.append(new_tiles)
+            return AGIPD_1M_SnappedGeometry(new_modules)
+        return self._snapped_cache
+
+    def position_modules_fast(self, data):
```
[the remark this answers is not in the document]
takluyver: Sure, added it back as a deprecated alias for `position_modules_fast`.

antarcticrainforest:
LGTM

PR #82, opened 2018-10-01
Value error raised in stack detector function fixed

ebadkamil:
Assuming that 'image.data' is of form (num_pulses, px,py). It was happening that some module sends image with num_pulses = 0 and   remaining modules num_pulses=60, then Value error was raised in Stack_detector_data function. I now fill those modules with num_pulses=0 to np.nan and they are now of shape of modules which do have full data for eg. (60,px,py).

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -1134,17 +1134,21 @@ def stack_detector_data(train, data, axis=-3, modules=16, only='', xcept=()):
         shapes.add(array.shape)
         modno_arrays[modno] = array
 
+    required_shape = (sorted(shapes, key=lambda x: x[0], reverse = True))[0]
     if len(dtypes) > 1:
         raise ValueError("Arrays have mismatched dtypes: {}".format(dtypes))
     if len(shapes) > 1:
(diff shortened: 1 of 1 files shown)


karabo_data/reader.py
```diff
@@ -1134,17 +1134,21 @@ def stack_detector_data(train, data, axis=-3, modules=16, only='', xcept=()):
...
         modno_arrays[modno] = array
 
+    required_shape = (sorted(shapes, key=lambda x: x[0], reverse = True))[0]
     if len(dtypes) > 1:
         raise ValueError("Arrays have mismatched dtypes: {}".format(dtypes))
     if len(shapes) > 1:
-        raise ValueError("Arrays have mismatched shapes: {}".format(shapes))
+        for modno, array in modno_arrays.items():
+            if array.shape != required_shape:
+                modno_arrays[modno] = np.full(required_shape, np.nan, dtype = array.dtype)
```
tmichela: You don't need to create a new array for all misshaped modules as an array of the right size is already made, line 1152

PR #84, opened 2018-10-09
Handle detector modules without data

### LINKED ISSUE #80: Value Error raised in stack_detector_data
When streaming from files for "AzimuthalIntegration FXE-tool" it happens for certain TrainId that some module has 'image.data'  of shape (0,256,256) which  raises Value Error in stack_detector_data function. Should Stack_detector_data function handle modules missing data?

tmichela:
Fixes #80


ebadkamil:
LGTM. Thanks.

PR #86, opened 2018-10-15
Allow selecting a list of trains

takluyver:
From discussion with @CFGrote on issue #85.

Open questions about out-of-range behaviour:

- Selecting a list of train IDs will currently always succeed, ignoring any IDs that aren't present in the source collection. We could strictly check that they are all in the source collection, or have a weaker check that at least one train matched. The latter is more like what we do for slicing by train ID: we throw a ValueError if the slice range doesn't overlap with the train IDs in the source collection.
- Selecting a list of indexes will fail if any of them are out of range. We could relax that.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -789,6 +789,9 @@ class DataCollection:
         elif isinstance(tr, by_id) and isinstance(tr.value, (list, np.ndarray)):
             # Select a list of trains by train ID
             new_train_ids = sorted(set(self.train_ids).intersection(tr.value))
+            if not new_train_ids:
+                raise ValueError("Given train IDs not found among {} trains in "
+                                 "collection".format(len(self.train_ids)))
         elif isinstance(tr, by_index) and isinstance(tr.value, (list, np.ndarray)):
             # Select a list of trains by index in this collection
             new_train_ids = sorted([self.train_ids[i] for i in tr.value])
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -761,7 +761,11 @@ class DataCollection:
             from karabo_data import by_id
             sel = run.select_trains(by_id[142844490:142844495])
 
-        Or by index within this collection::
+        Or select a list of trains::
+
+            sel = run.select_trains(by_id[[142844490, 142844493, 142844494]])
+
+        Or select trains by index within this collection::
 
             from karabo_data import by_index
             sel = run.select_trains(by_index[:5])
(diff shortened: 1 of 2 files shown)


takluyver:
I've implemented the weaker check: error when selecting a list of train IDs if none of them match. I'm not convinced that this is the right behaviour, though.

CFGrote:
from the practical point of view, this is our desired behavior.

karabo_data/reader.py
```diff
@@ -773,16 +777,27 @@ def select_trains(self, train_range):
...
+        tr = train_range
+        if isinstance(tr, by_id) and isinstance(tr.value, slice):
+            # Slice by train IDs
+            start_ix = _tid_to_slice_ix(tr.value.start, self.train_ids, stop=False)
+            stop_ix = _tid_to_slice_ix(tr.value.stop, self.train_ids, stop=True)
+            new_train_ids = self.train_ids[start_ix:stop_ix:tr.value.step]
+        elif isinstance(tr, by_index) and isinstance(tr.value, slice):
+            # Slice by indexes in this collection
+            new_train_ids = self.train_ids[tr.value]
+        elif isinstance(tr, by_id) and isinstance(tr.value, (list, np.ndarray)):
```
tmichela: Can we allow any iterable? `instance(tr.value, collections.Iterable)`

takluyver: I'm reluctant to accept any iterable because so many things (e.g. strings) are iterable, even when using them wouldn't make sense. Or you could construct e.g. `range(0, int(1e10))`, which will work, but with terrible performance.

Maybe it would make sense to check what types numpy accepts for indexing arrays.

takluyver: > NumPy arrays may be indexed with other arrays (or any other sequence- like object that can be converted to an array, such as lists, with the exception of tuples; see the end of this document for why this is).

So maybe we should be more flexible about this.

A sequence is an ordered collection; karabo_data will currently ignore the order of the indexes given and keep the trains in sequential order. So it may make sense to index by an unordered collection like a set.

takluyver: Out of pragmatism, I think we should leave it to lists and arrays for now: those are the most likely types for people to pass, it's easy to convert another sequence to them, and it avoids any complications with unexpected types matching a broader check.

We can always broaden it later if the limitation turns out to be an inconvenience, but I'd rather make that a practical decision (what do people want to use?) rather than an abstract one.

tmichela:
LGTM

PR #88, opened 2018-10-23
Make run sources frozen sets

cydanil:
Previously, a user could mess up known sources by popping from the sources sets:
```python
In [1]: from karabo_data import RunDirectory
In [2]: run = RunDirectory(".")

In [3]: run.all_sources
Out[3]: {'SQS_DIGITIZER_UTC1/ADC/1', 'SQS_DIGITIZER_UTC1/ADC/1:network'}

In [4]: run.instrument_sources.pop()
Out[4]: 'SQS_DIGITIZER_UTC1/ADC/1:network'

In [5]: run.all_sources
Out[5]: {'SQS_DIGITIZER_UTC1/ADC/1'}
```
All four `{control,instrument,detector,all}_sources` were affected

This resolves this issues by making these frozensets.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -163,6 +163,9 @@ class FileAccess:
             else:
                 raise ValueError("Unknown data category %r" % category)
 
+        self.control_sources = frozenset(self.control_sources)
+        self.instrument_sources = frozenset(self.instrument_sources)
+
         # {(file, source, group): (firsts, counts)}
         self._index_cache = {}
         # {source: set(keys)}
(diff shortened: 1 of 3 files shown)


takluyver:
Thanks, makes sense. Can you add a test in `test_reader_mockdata` so it can run without relying on data in @tmichela's home directory?

(We should probably get rid of the `test_reader` module at some point - I think `test_reader_mockdata` is now more complete anyway)

cydanil:
Thanks for your input! I then noticed I had not made all the required changes.
I've added tests to `test_reader_mockdata`

takluyver:
Thanks :-)

PR #96, opened 2018-11-20
Compare different AGIPD geometries

takluyver:
This work-in-progress branch adds a method for graphically comparing different geometries, so you can see how they've changed.

example comparison plot

### DIFF
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -100,8 +100,11 @@ class AGIPD_1MGeometry:
     of the pixel size.
     """
     pixel_size = 2e-7  # 2e-7 metres == 0.2 mm
-    def __init__(self, modules):
+    def __init__(self, modules, filename='No file'):
         self.modules = modules  # List of 16 lists of 8 fragments
+        # self.filename is metadata for plots, we don't read/write the file.
+        # There are separate methods for reading and writing.
+        self.filename = filename
         self._snapped_cache = None
 
     @classmethod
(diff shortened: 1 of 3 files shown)


takluyver:
I think this is ready for review.

takluyver:
Ping @tmichela @ebadkamil - this is only about 100 lines of new code, and it also fixes a dependency problem from cfelpyutils that is making the tests fail on Travis.

karabo_data/geometry2.py
```diff
@@ -147,7 +150,7 @@ def from_crystfel_geom(cls, filename):
             for a in range(8):
                 d = geom_dict['panels']['p{}a{}'.format(p, a)]
                 tiles.append(AGIPDGeometryFragment.from_panel_dict(d))
-        return cls(modules)
+        return cls(modules,  filename=filename)
```
tmichela: One extra space character

tmichela:
LGTM, thanks

karabo_data/geometry2.py
```diff
@@ -99,8 +99,11 @@ class AGIPD_1MGeometry:
     of the pixel size.
     """
     pixel_size = 2e-7  # 2e-7 metres == 0.2 mm
```
ebadkamil: pixel_size = 2e-4

karabo_data/geometry2.py
takluyver: Eagle eyes :-P

Fixed

karabo_data/geometry2.py
takluyver: Yup, but you're already on that in PR #97 :+1:

PR #97, opened 2018-11-21
Distortion

ebadkamil:
Generate distortion for LPD and AGIPD detectors

From given geometries for LPD and AGIPD, now we can calculate distortion array of shape (Ny=1024,Nx=1024,Nc=4,3) which can be used for defining detectors in Nexus format (pyFAI compliant too).

### DIFF
--- a/karabo_data/geometry.py
+++ b/karabo_data/geometry.py
@@ -195,8 +195,8 @@ class LPDGeometry(GeometryFragment):
 
         Returns a matplotlib Figure object.
         """
-        # from matplotlib.backends.backend_agg import FigureCanvasAgg
-        # from matplotlib.figure import Figure
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+        from matplotlib.figure import Figure
 
         fig = Figure((10, 10))
         FigureCanvasAgg(fig)
@@ -231,31 +231,31 @@ class LPDGeometry(GeometryFragment):
         ax.set_title('LPD detector geometry')
         return fig
 
-    def generateDistortion(self):
+    def to_distortion_array(self):
         """ Return distortion matrix for LPD detector
 
         Returns
         -------
         out: ndarray
-            Dimension (Ny=1024, Nx=1024, 4, 3)
+            Dimension (4096=16(modules)*256(ss_dim), 256(fs_dim), 4, 3)
             type: float32
             4 is the number of corners of a pixel
             last dimension is for Z, Y, X location of each corner
         """
-        _quad, _modules, _tiles = self._create_qmt_dict()
+        _tiles = self._create_qmt_dict()
         _, centre = self._plotting_dimensions()
-        distortion = np.zeros((1024, 1024, 4, 3), dtype=np.float32)
+        distortion = np.zeros((4096, 256, 4, 3), dtype=np.float32)
 
-        for (Q, M) in product(range(1, 5), range(1, 5)):
-
+        for index, (Q, M) in enumerate(product(range(1, 5), range(1, 5))):
+            #ch: module offset along first dimension of distortion array
+            ch = index*256
             for T in range(1, 17):
                 position = self.find_offset(('Q%d' % Q,
                                              'M%d' % M, 'T%02d' % T))
 
-                # idx, idy : Bottom left index of tiles in (1024,1024) array
-                idx, idy = np.add(_quad['Q%d' % Q],
-                                  np.add(_modules['M%d' % M],
-                                  _tiles['T%01d' % T]))
+                # idx, idy : Bottom left index of tiles in (256,256)
+                #            module array
+                idx, idy = _tiles['T%01d' % T]
 
                 xt, yt = (position // self.pixel_size) + centre
                 xt, yt = int(xt), int(yt)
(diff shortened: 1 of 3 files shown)


takluyver:
I think it would be better to use the distortion (and corresponding data) as a 8192 x 128 array (for AGIPD) or 4096 x 256 (for LPD) rather than 1024 x 1024.

- Reliability: it's not obvious which bit of a 1024 x 1024 array is which module. Do we try to emulate the detector layout, which is a strange order based on how the hardware is built, or do we lay them out in 'reading order'? The long strip arrangement makes it clear that we're not trying to emulate the detector layout, and it's just counting up from 0 to 15.
- Efficiency: assembling a 1024 x 1024 array means the pixels from different modules are interleaved row by row in memory. Laying them out in a single long strip means each module is one chunk of memory. It's the same memory layout as when we stack the data.

karabo_data/geometry2.py
```diff
@@ -98,7 +99,7 @@ class AGIPD_1MGeometry:
     The coordinates used in this class are 3D (x, y, z), and represent multiples
     of the pixel size.
     """
-    pixel_size = 2e-7  # 2e-7 metres == 0.2 mm
+    pixel_size = 2e-4  # 2e-4 metres == 0.2 mm
```
takluyver: :+1:

karabo_data/geometry2.py
```diff
@@ -336,6 +337,116 @@ def plot_data_fast(self, data):
         """
         return self._snapped().plot_data(data)
 
+    def generateDistortion(self):
```
takluyver: For consistency, let's call these something like `to_distortion_array`.

karabo_data/geometry2.py
```diff
@@ -336,6 +337,116 @@ def plot_data_fast(self, data):
...
+        """ Return distortion matrix for LPD detector
```
takluyver: AGIPD

karabo_data/geometry.py
```diff
@@ -231,11 +231,96 @@ def inspect(self):
...
+                    self.pixel_size*y_position[:, :] + 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 3, 1] = (
+                    self.pixel_size*y_position[:, :] - 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 0, 2] = (
+                    self.pixel_size*x_position[:, :] - 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 1, 2] = (
+                    self.pixel_size*x_position[:, :] - 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 2, 2] = (
+                    self.pixel_size*x_position[:, :] + 0.5*self.pixel_size)
+                distortion[idy:idy+32, idx:idx+128, 3, 2] = (
```
takluyver: Does this not need the same `ch+idy` as the others?

karabo_data/geometry.py
```diff
@@ -231,11 +231,96 @@ def inspect(self):
...
+                # of its 4 corners in real space.
+                # Z position is set to 0 (Flat detector)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 0, 1] = (
+                    self.pixel_size*y_position[:, :] - 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 1, 1] = (
+                    self.pixel_size*y_position[:, :] + 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 2, 1] = (
+                    self.pixel_size*y_position[:, :] + 0.5*self.pixel_size)
+                distortion[ch+idy:ch+idy+32, idx:idx+128, 3, 1] = (
+                    self.pixel_size*y_position[:, :] - 0.5*self.pixel_size)
```
takluyver: I think it should be possible to do this with numpy broadcasting, rather than assigning the four corners separately, something like this (untested):

```python
corner_y_offsets = np.array([-.5, .5, .5, -.5]) * self.pixel_size

distortion[ch+idy:ch+idy+32, idx:idx+128, :, 1] = (
                    self.pixel_size*y_position[:, :, np.newaxis] + corner_y_offsets)
```

karabo_data/geometry.py
```diff
@@ -231,11 +231,96 @@ def inspect(self):
...
+                # idx, idy : Bottom left index of tiles in (256,256)
+                #            module array
+                idx, idy = _tiles['T%01d' % T]
+
+                xt, yt = (position // self.pixel_size) + centre
+                xt, yt = int(xt), int(yt)
+                # y_position contains y_indices of pixels for a tile
+                y_position = np.array(
+                    [yt+j for j, _ in product(range(32), range(128))]).\
+                    reshape(32, 128)
```
takluyver: I think [np.meshgrid](https://docs.scipy.org/doc/numpy/reference/generated/numpy.meshgrid.html) might be useful for these.

karabo_data/geometry.py
ebadkamil: done. thanks

karabo_data/geometry.py
ebadkamil: done.

ebadkamil:
LPD distortion produces correct image when assembly is done using pyFAI Distortion now. 
Let's discuss about AGIPD distortion tomorrow in group meeting. It's confusing to me.

takluyver:
Can I clarify anything about the AGIPD geometry? Or I can have a go at the code myself, if you prefer.

ebadkamil:
Hi Thomas. It will great if you can check the AGIPD distortion (geometry2.py) and change as you wish.

takluyver:
Thanks, just rebased it onto master because there was a merge conflict. I'll have a look now.

Do you have a script or something handy to apply a distortion array to some data, so I can test the arrays produced?

takluyver:
I'm having some trouble with the coordinate system for LPD. I've just sent an email about it. @ebadkamil has worked around the coordinate question by flipping the data for each module before passing it to pyFAI, but I'd like to work out the coordinates so that's unnecessary.

For AGIPD, it now seems to be working well:

agipd_distortion_1

takluyver:
I suggest that we either merge this as is, or separate the distortion array for AGIPD (which I'm happy with) and leave LPD for now. If we can get the coordinate system clarified, I'd like to try to build new code for LPD geometry on a similar basis to the AGIPD code.

@ebadkamil @tmichela @fangohr how do we want to handle review & decisions on this PR, which Ebad started and then I made further changes to?

tmichela:
If the issue with LPD clarified already? if not I would merge the AGIPD implementation already

fangohr:
Ideally a third person reviews. Alternatively, @ebadkamil can review your changes and vice versa.

takluyver:
#103 is just the AGIPD changes, if people want to review that. Once we know where we are with coordinates, I'd like to build new geometry code for LPD on the same basis as the AGIPD geometry.

ebadkamil:
I will talk to Steffen regarding LPD geometry (to clarify few points for the presentation for the user workshop) this week provided he has time. Thomas M will join too. We should clarify why one needs to invert tiles array in both axis.

ebadkamil:
@takluyver  aren't we inverting data on both axis?
Line 133

`                out[..., y0:y0 + tile_data.shape[-2],
                         x0:x0 + tile_data.shape[-1]] = tile_data[..., ::-1, ::-1]`

takluyver:
Yes, looking at it, I think the x coordinates we're using are also reversed. Please do check this with Steffen as well.

I'm referring for this to the [examining detector geometry](https://karabo-data.readthedocs.io/en/latest/examine_geometry.html) page in the docs. You can see the (x, y) coordinates of the quadrants we got from Steffen based on positions in March, along with the layout matching the [LPD layout diagram](https://in.xfel.eu/readthedocs/docs/detector-documentation/en/latest/lpd/source/introduction.html).

It's more fiddly for the x axis, because flipping it is the difference between a perspective looking 'into the beam' (like a camera) versus a 'beam's eye view' onto the detector. However, the LPD layout diagram says it is 'always looking in beam direction', which I assume means onto the detector. That should mean that increasing x in the XFEL coordinate system is to the left in this diagram, whereas we're positioning the quadrants with x increasing to the right.

Frustratingly, this display convention is the opposite to CrystFEL, which has x increasing to the right, or an 'into the beam' view. We might need to write our code with a switch for displaying images for people who're used to each convention.

tmichela:
I believe this is outdated? should we close?

takluyver:
Yep, it can. We'll do distortion arrays for LPD based on the geometry2 code.

Issue #101, opened 2018-12-19
Can't read image data anymore


hjkirkwood:
I went to re-run some code using karabo_data.RunDirectory from ~ 2 months ago and ran into a problem, below is some code to reproduce it.  I can read the data if I use h5py so I believe the data is fine (I'm reading from "RAW" files) but this is much less convenient.

The karabo `run.info()` works fine, but when I call `run.get_array(device,key)` I run into trouble.

```
import karabo_data
karabo_data.__version__
#this gives: '0.2.0'


from karabo_data import RunDirectory

topdir = '/pnfs/xfel.eu/exfel/archive/XFEL/raw/SPB/201701/p002046'
# generator for run directory for karabo_data.RunDirectory
run_pattern = lambda nRun: topdir + '/r{:04d}'.format(int(nRun))

# get dark run data (I tested this multiple runs and get the same result)
run_dark = RunDirectory(run_pattern(85))


device = 'SPB_EHC_PAM/CAM/SPEC12:daqOutput'
key = 'data.image.pixels'

# get run images for above device
dark_ims = run_dark.get_array(device,key)
```

I get the following traceback:

```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
<ipython-input-2-775b2bc84686> in <module>()
      3 
      4 # get run images for above device
----> 5 dark_ims = run_dark.get_array(device,key)

~/karabo_data/karabo_data/reader.py in get_array(self, source, key, extra_dims)
    630                                    key=lambda a: a.coords['trainId'][0]),
    631                             dim='trainId')
--> 632         return arr.sel(trainId=self.train_ids)
    633 
    634     def union(self, *others):

~/.local/lib/python3.6/site-packages/xarray/core/dataarray.py in sel(self, indexers, method, tolerance, drop, **indexers_kwargs)
    782         indexers = either_dict_or_kwargs(indexers, indexers_kwargs, 'sel')
    783         ds = self._to_temp_dataset().sel(
--> 784             indexers=indexers, drop=drop, method=method, tolerance=tolerance)
    785         return self._from_temp_dataset(ds)
    786 

~/.local/lib/python3.6/site-packages/xarray/core/dataset.py in sel(self, indexers, method, tolerance, drop, **indexers_kwargs)
   1507         indexers = either_dict_or_kwargs(indexers, indexers_kwargs, 'sel')
   1508         pos_indexers, new_indexes = remap_label_indexers(
[...]

PR #102, opened 2018-12-19
Use intersection of selected & available trains for get_array/get_series

takluyver:
Closes gh-101

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -535,7 +535,7 @@ class DataCollection:
             raise Exception("Unknown source category")
 
         ser = pd.concat(sorted(seq_series, key=lambda s: s.index[0]))
-        return ser.loc[self.train_ids]
+        return ser.loc[ser.index.intersection(self.train_ids)]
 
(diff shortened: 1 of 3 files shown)


takluyver:
Looks like I've broken something else for pandas. At least our tests caught that one. I'll investigate tomorrow.

takluyver:
I fixed the second problem, but I still want to add a test case for the first problem (described in #101) before we merge it.

@hjkirkwood I'm assuming you can use this branch for a little while to do what you need, but let us know if you need the fix merged into master sooner.

takluyver:
I got round to adding the test I wanted on the plane, so this is ready for review now.

I also noticed some deprecation warnings from h5py, which were easy to fix, so I've done that in this PR.

tmichela:
LGTM,

minor: would be nice to have a test that select a data source with missing `trainIds` (not containing data) for which other data sources have data.

tmichela:
Ah, my bad, I did not get that the data-set creation would skip every second train,
LGTM!

takluyver:
No problem, that's not at all obvious from the changes in this pull request.

PR #103, opened 2019-01-02
AGIPD geometry to distortion array

takluyver:
Based on #97, picking only the changes for AGIPD, not LPD, while we work out the coordinates issue for LPD better.


tmichela:
LGTM

PR #107, opened 2019-02-05
Start Integrating LPD in geometry2 module

takluyver:
`karabo_data.geometry2` is my second attempt at dealing with detector geometry, initially written with AGIPD geometry. Compared to `karabo_data.geometry`, it can store data about 3D geometry, including modules tilted and rotated. The methods we mostly use to assemble images 'snap' the geometry to a 2D, axis-aligned grid to work more efficiently, but more accurately representing the geometry is important where precision matters, e.g. for finding peaks for crystallography.

This starts to generalise the `geometry2` machinery to handle different detectors, and in particular LPD. So far, the methods available are:

- `from_quad_positions()` to create an approximate geometry from a few numbers.
- `inspect()` to show a diagram of the geometry
- `position_modules_fast()` / `position_all_modules()`
- `plot_data_fast()`

The main things still to be generalised are I/O: reading and writing CrystFEL and XFEL format geometry files, and creating distortion arrays for PyFAI. I could also generalise the assembly by affine transform, but it's so slow it's probably not useful.

I thought I'd stop at this point for code review while the changes are still hopefully a manageable size.

Here's an LPD image assembled by the new machinery:

screenshot from 2019-02-05 12-37-49

It's flipped left-right compared to the images from the older geometry machinery. This is a consequence of following the XFEL coordinate system. I may look at adding a switch to flip the plot back.


tmichela:
maybe @ebadkamil and @antarcticrainforest can have a look here

takluyver:
I've just improved the plotting, for both AGIPD and LPD.

- The 0 of each axis is now at the nominal reference point in the centre of the detector, not at the corner.
- The axes are labelled with pixels by default, but you can switch that to metres with `axis_units='m'`.
- You can flip the x-axis left-right to get the image looking from the beam direction by setting `frontview=True`. This view appears to be the convention in Karabo, although it doesn't directly match the coordinate system.

I'm open to other ideas for the naming and defaults of the two new parameters. I initially called the second `fliplr`, but I don't like the reliance on context: to know what flipping does, you have to know what it shows normally.

Here's what it looks like with the new options in use:

screenshot from 2019-02-11 12-50-12

takluyver:
I've now added reading XFEL geometry files (so `geometry2` now has all the functionality for LPD that `geometry` does), and updated the example notebooks.

I've created an inverted version of the geometry file we got from DET, so that the coordinates work in the XFEL standard coordinate scheme. I'm assuming that the coordinates still point to the last pixel of each part (i.e. the corner by M4 for each quadrant, and the corner by T16 for each module), and that they're still in mm. I'd like to check both of these things with Steffen.

At present the quadrant positions you use with the XFEL geometry files (mm to M4 corner) are also completely different from the quadrant positions you use to generate the idealised geometry without a file (pixels to M1 corner). Once I've confirmed the details with Steffen, I'll straighten this difference out.

antarcticrainforest:
Just one quick question. What will happen to the geometry module? Are you planning to rename geometry2 to geometry? I am asking because a few people got already confused.

karabo_data/geometry2.py
```diff
@@ -745,6 +754,66 @@ def from_quad_positions(cls, quad_pos, asic_gap=4, panel_gap=4):
                 ))
         return cls(modules)
 
+    @classmethod
+    def from_h5_file_and_quad_positions(cls, path, positions, unit=1e-3):
```
antarcticrainforest: Are you also planning to add a write method?

takluyver:
I was planning to leave the geometry module as is for the moment, so we don't break existing code, but encourage people to move to geometry2. Then we could eventually remove geometry after giving people some time to switch.

A more aggressive approach would be to replace the geometry module directly. I've tried to keep the API similar on the new implementation. But one crucial difference is the XFEL geometry file format: as we discussed with Steffen, I flipped the coordinates to match the XFEL standard coordinate system. The new code expects them flipped, so using it with the old file is going to give weird results.

karabo_data/geometry2.py
takluyver: Yes, but probably not in this pull request; I'm trying not to make it too big a change to review.

takluyver:
Confirmed with Steffen that they plan to keep the same reference points and units in the geometry files, so I've updated `LPD_1MGeometry.from_quad_positions()` to use this scheme (mm to M4 corner).

This now means that the quadrant positions used for LPD (mm to last pixel corner) are different from those for AGIPD (pixels to first pixel corner). I've added a `unit=` parameter to both to make it easier to switch between mm and pixels. I don't have a good answer for the difference in corners, but since the quadrant positions are numbers that you copy from somewhere, I don't think it's too big a problem that they are different for different detectors, so long as they're consistent for a given detector.

karabo_data/geometry2.py
```diff
@@ -631,6 +695,229 @@ def plot_data(self, modules_data):
...
+        panel_gap_px = 4 if (panel_gap is None) else panel_gap * px_conversion
+
+        tile_size = np.array([cls.frag_fs_pixels, cls.frag_ss_pixels, 0])
+
+        panels_across = [-1, -1, 0, 0]
+        panels_up = [0, -1, -1, 0]
+        modules = []
+        for p in range(16):
+            quad = p // 4
+            quad_corner_x = quad_pos[quad][0] * px_conversion
```
ebadkamil: This quad_position is wrt to centre? Is the hole in detector the centre?

ebadkamil: Me and Martin were trying to understand the quad positions we use as of now

- quad_pos = [(-11.4, -299), (11.5, -8), 
                (-254.5, 16), (-278.5, -275)]

In the docstring you wrote "In the initial detector layout, the corner positions are for the top right corner of the quadrant, looking into the beam."  Somehow this doesn't look consistent if the centre of detector is the origin and y axis points up, x axis points left and z axis point into the detector center.

karabo_data/geometry2.py
```diff
@@ -631,6 +695,229 @@ def plot_data(self, modules_data):
...
+        The quadrant positions refer to the corner of each quadrant
+        where module 4, tile 16 is positioned.
+        This is the corner of the last pixel as the data is stored.
+        In the initial detector layout, the corner positions are for the top
+        right corner of the quadrant, looking into the beam.
+
+        Parameters
+        ----------
+        quad_pos: list of 2-tuples
+          (x, y) coordinates of the last corner (the one by module 4) of each
```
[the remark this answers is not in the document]
takluyver: I want to identify which corner of the quadrant you're giving the position for without referring to left and right, because they depend on which way you plot the image. But in [the diagram here](https://in.xfel.eu/readthedocs/docs/detector-documentation/en/latest/lpd/source/introduction.html), for instance, they're the top left corners of each quadrant.

Each LPD quadrant has 4 (super)modules, numbered 1-4. Whichever way you plot the image, each quadrant is positioned by its corner nearest to module 4 in that quadrant.

I'm scribbling lots of little diagrams on pieces of paper as I work on this to make sense of what's where. It's really hard to describe clearly. I'm starting to think we should borrow the terms 'port' and 'starboard', because sailors solved exactly the same problem with left/right centuries ago.

karabo_data/geometry2.py
takluyver: Yup, I should make it clear that the origin is in the centre of the detector. I think the idea is that it's the point where the beam intersects with the detector plane, but I imagine you have to do some refinement if you want a precise beam centre.

The quadrant positions we're using at the moment are backwards (x axis beam-right, y axis down), like the coordinates in the HDF5 file. The new machinery takes coordinates in the standard directions (x axis beam-left, y axis up), so all the numbers have to be inverted. I've done that in the examples:

https://github.com/European-XFEL/karabo_data/pull/107/files#diff-90dbb13e16d6833f559126522a31bc42R46

karabo_data/geometry2.py
```diff
@@ -216,13 +221,15 @@ def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
             x_orient = quads_x_orientation[quad]
             y_orient = quads_y_orientation[quad]
             p_in_quad = p % 4
-            corner_y = quad_corner[1] - (p_in_quad * (128 + panel_gap))
+            corner_y = (quad_corner[1] * px_conversion)\
```
antarcticrainforest: Stupid question: should those values be rounded to int, since they're representing pixels

karabo_data/geometry2.py
```diff
@@ -216,13 +221,15 @@ def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
...
+                       - (p_in_quad * (128 + panel_gap))
 
             tiles = []
             modules.append(tiles)
 
             for a in range(8):
-                corner_x = quad_corner[0] + x_orient * (64 + asic_gap) * a
+                corner_x = (quad_corner[0] * px_conversion)\
+                           + x_orient * (64 + asic_gap) * a
```
antarcticrainforest: Can you remind me again where the 64 comes from?

karabo_data/geometry2.py
takluyver: They're representing pixel units, but we want to maintain sub-pixel accuracy for these positions. We round them for plotting (that's the 'snapped' geometry), but the more accurate positions can still be useful for CrystFEL and PyFAI.

karabo_data/geometry2.py
takluyver: That's the ASIC/tile size in pixels (this bit of the code is for AGIPD). There's a named value that it should be using here; I'll fix that.

karabo_data/geometry2.py
antarcticrainforest: Are you sure that quad_pos in the notebook are still correct? We were trying to reconstruct the quad positions based on these coordinates and it does not add up.

antarcticrainforest: looking at geom.inspect shows that the quadrants starting form the top left and going counter clockwise to the top right, but according to the detector intro https://in.xfel.eu/readthedocs/docs/detector-documentation/en/latest/lpd/source/introduction.html the quadrants should start from the top right going clockwise to the to left

antarcticrainforest: Dude, when will you be here? I guess Ebad an me are too stupid. Maybe you can show us once your back in HH?

takluyver: The diagram in the detector intro is shown in 'beam direction', i.e. as if you were standing with the beam going through you, looking at the detector. `geometry.inspect()` plots it the other way, as if you were looking into the beam (this way follows the usual convention of plotting x increasing left to right). So left/right and clockwise/anticlockwise are flipped.

This makes my head hurt too. :wink: 

For the `plot_data_fast` function, I added a `frontview=True` parameter to flip the image to match the orientation in the diagram ([details above](https://github.com/European-XFEL/karabo_data/pull/107#issuecomment-462318003)). I was going to do the same for `inspect()` in a separate PR to keep the changes here smaller, but if you'd find it easier to review with that, I can add it to this PR.

takluyver:
I've written some API documentation for this, including diagrams which I hope will make it clearer what I'm talking about.

I've also tried to standardise both the `inspect()` method and `plot_data_fast()` to show a 'front view' (looking along the beam) by default. This means that the x axis is 'backwards' (increasing right to left), but it matches the way Karabo displays images, and the diagrams from the detector group, so hopefully this means less mental effort switching between two different views.

You can pass `frontview=False` if you prefer to see the images as if you were looking into the beam.

karabo_data/geometry2.py
zhujun98: I do not understand the benefit of adding a "center" for the detector since it differs the top-left-corner-origin convention of displaying an image. It increases the burden of our brains since we may need to digest the images with different origins at the same time. 

In reality, the beam usually does not pass through exactly the detector center. Therefore, having a beam position of (2, 3) is not more convenient than having a beam position of (602, 603). In addition, in some experiments, the beam does not even pass through the hole (it bypasses the detector). 

Moreover, speaking of pyFAI, it adopts the bottom-left-corner-origin convention. It means that we may be facing three different origins at the same time :(

takluyver: The CrystFEL geometry already seems to use the centre as the origin, as do the quadrant positions we got to use with the XFEL geometry files. So it's not a new convention we're adding.

An 'image style' origin in the top left doesn't make sense given XFEL's standard coordinate system, because we have y increasing upwards and x increasing beam-left.

zhujun98: Nevertheless, the Karabo scenes always display images with the origin being the top left corner. XFEL / CrystFEL's conventions are both implementation details which users may not care. The displayed image is the interface that affects the users. Of course, one could argue CrystFEL is the user. Generally, for people who is doing data analysis, they received the image as an array, and the array index starts from 0.

takluyver: Displaying images is only one use case: you might also save the geometry to a CrystFEL file, or convert it to a PyFAI distortion array. Or an XFEL file, once that's implemented. Shifting the origin to the topmost/leftmost pixels throws away information about where the centre is, which we need in order to write out geometry files comparable to those we read.

Also, if you define the origin in line with the topmost/leftmost pixels, then moving the topmost or leftmost panels in the geometry means moving all the other panels, because the origin shifts. Obviously that's possible if there's no other choice, but it's not elegant, and it makes comparing different geometries harder.

zhujun98: Thanks for the explanation! In principle, displaying images can be totally decoupled from saving geometry, converting to something else or any other internal manipulations. But if it is an expert interface, then it makes sense to also display the implemented image coordinate system.

karabo_data/geometry2.py
```diff
@@ -631,6 +737,228 @@ def plot_data(self, modules_data):
...
+            quad_corner_y = quad_pos[quad][1] * px_conversion
+
+            p_in_quad = p % 4
+            # Top beam-left corner of panel
+            panel_corner_x = (quad_corner_x +
+                              (panels_across[p_in_quad] * (
+                                          256 + asic_gap_px + panel_gap_px)))
+            panel_corner_y = (quad_corner_y +
+                              (panels_up[p_in_quad] * (256 + (
+                                          7 * asic_gap_px) + panel_gap_px)))
```
antarcticrainforest: Are the 256 coming from the ``expected_data_shape`` tuple? Could you also refer to  them by ``expected_data_shape[1/2]`` rather than writing 256 explicitly?

karabo_data/geometry2.py
```diff
@@ -631,6 +737,228 @@ def plot_data(self, modules_data):
...
+                ax.text(cx, cy, 'T{}'.format(t + 1),
+                        verticalalignment='center',
+                        horizontalalignment='center')
+
+        ax.set_title('LPD-1M detector geometry ({})'.format(self.filename))
+        return fig
+
+    @staticmethod
+    def split_tiles(module_data):
+        lhs, rhs = np.split(module_data, 2, axis=-1)
```
antarcticrainforest: As we've had discussed left and right are kind of arbitrary. Nautical terms port and starboard might also be confusing

karabo_data/geometry2.py
```diff
@@ -92,22 +99,142 @@ def __init__(self, corner_pos, ss_vec, fs_vec):
...
+        """Plot the 2D layout of this detector geometry.
+
+        Returns a matplotlib Figure object.
+        """
+        from matplotlib.backends.backend_agg import FigureCanvasAgg
+        from matplotlib.collections import PatchCollection
+        from matplotlib.figure import Figure
+        from matplotlib.patches import Polygon
+
+        fig = Figure((10, 10))
```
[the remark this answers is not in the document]
takluyver: Since you showed me the problems with using this outside a notebook yesterday, I've switched it over to creating a figure with pyplot.

karabo_data/geometry2.py
```diff
@@ -92,22 +99,142 @@ def __init__(self, corner_pos, ss_vec, fs_vec):
...
+
+        data : ndarray
+          Should have exactly 3 dimensions, for the modules, then the
+          slow scan and fast scan pixel dimensions.
+        axis_units : str
+          Show the detector scale in pixels ('px') or metres ('m').
+        frontview : bool
+          If True (the default), x increases to the left, as if you were looking
+          along the beam. False gives a 'looking into the beam' view.
+        """
```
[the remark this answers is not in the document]
takluyver: This method calls `SnappedGeometry.plot_data`, which does check that.

karabo_data/geometry2.py
```diff
@@ -116,7 +243,18 @@ def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
         The quadrant positions are given in pixel units, referring to the first
         pixel of the first module in each quadrant, corresponding to data
         channels 0, 4, 8 and 12.
+
+        The origin of the coordinates is in the centre of the detector.
+        Coordinates increase upwards and to the left (looking along the beam).
+
+        To give positions in different units, pass the *unit* parameter as the
+        length of the unit in metres. E.g. ``unit=1e-3`` means the coordinates
```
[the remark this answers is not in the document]
takluyver: Responding to the similar comment below

karabo_data/geometry2.py
```diff
@@ -116,7 +243,18 @@ def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
...
+        are in millimetres.
         """
+        px_conversion = unit / cls.pixel_size
```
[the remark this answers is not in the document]
takluyver: This unit doesn't describe the pixel size (which is fixed for a given detector), but rather what units the quadrant positions (and `asic_gap` and `panel_gap`) are given in. By default they're in pixel units, i.e. 1 means 0.2mm, the side length of 1 pixel, thus `unit=2e-4`. If you prepare the positions in mm, you would use `unit=1e-3` instead.

This might not be used much, but I added it so that the constructors for AGIPD and LPD are a bit more similar, with different units by default instead of hardcoded different units.

I'm adjusting the wording slightly, I hope it will make it clearer.

karabo_data/geometry2.py
takluyver: Kind of, but `expected_data_shape` relates to the data we get, rather than the physical layout of the detector itself. I've had a go at clarifying this in a different way.

karabo_data/geometry2.py
```diff
@@ -58,16 +63,18 @@ def snap(self):
...
+class GridGeometryFragment:
     # These coordinates are all (y, x), suitable for indexing a numpy array.
-    def __init__(self, corner_pos, ss_vec, fs_vec):
+    def __init__(self, corner_pos, ss_vec, fs_vec, ss_pixels, fs_pixels):
         self.ss_vec = ss_vec
         self.fs_vec = fs_vec
+        self.ss_pixels = ss_pixels
+        self.fs_pixels = fs_pixels
+
         if fs_vec[0] == 0:
```
ebadkamil: Isn't fs_vec[0] is 0 always ?

`assert {tuple(np.abs(ss_vec)), tuple(np.abs(fs_vec))} == {(0, 1), (1, 0)}`

`return GridGeometryFragment(corner_pos[::-1], ss_vec[::-1], fs_vec[::-1], self.ss_pixels, self.fs_pixels)`

karabo_data/geometry2.py
```diff
@@ -58,16 +63,18 @@ def snap(self):
...
         # Convert xy coordinates to yx indexes
-        return GridGeometryFragment(corner_pos[::-1], ss_vec[::-1], fs_vec[::-1])
+        return GridGeometryFragment(corner_pos[::-1], ss_vec[::-1], fs_vec[::-1],
+                                    self.ss_pixels, self.fs_pixels)
 
-class GridGeometryFragment:
-    ss_pixels = 64
-    fs_pixels = 128
 
+class GridGeometryFragment:
```
ebadkamil: This class is important for snap/ideal geometry. It must have a docstring especially explaining flip without transpose or transpose then flip part.

karabo_data/geometry2.py
```diff
@@ -116,7 +241,18 @@ def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29):
...
+
+        To give positions in units other than pixels, pass the *unit* parameter
+        as the length of the unit in metres.
+        E.g. ``unit=1e-3`` means the coordinates are in millimetres.
         """
+        px_conversion = unit / cls.pixel_size
+        asic_gap *= px_conversion
+        panel_gap *= px_conversion
+
         quads_x_orientation = [1, 1, -1, -1]
```
ebadkamil: Sorry to bring this again but in this layout 
https://github.com/European-XFEL/karabo_data/blob/lpd-geom2/docs/_static/agipd_layout.png 

x orientation of Quadrant 1 and 2 is in negative X direction. Or `quads_x_orientation` denotes something else here?

karabo_data/geometry2.py
```diff
@@ -631,6 +737,228 @@ def plot_data(self, modules_data):
...
+    You won't normally instantiate this class directly:
+    use one of the constructor class methods to create or load a geometry.
+    """
+    pixel_size = 5e-4  # 5e-4 metres == 0.5 mm
+    frag_ss_pixels = 32
+    frag_fs_pixels = 128
+    expected_data_shape = (16, 256, 256)
+
+    @classmethod
+    def from_quad_positions(cls, quad_pos, *, unit=1e-3, asic_gap=None,
```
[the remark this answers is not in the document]
takluyver: The idea is that `asic_gap` and `panel_gap`, if you pass them, are also in the specified units (so mm by default). So 4 would mean 4mm, which is 8 LPD pixels. Having numbers in the defaults then also means that those gaps change if you change the units without specifying the gaps.

I'm torn because I like having the numbers in the defaults as well. I actually did leave them in for the equivalent AGIPD method, where the default units are multiples of pixel size.

karabo_data/geometry2.py
```diff
@@ -631,6 +737,228 @@ def plot_data(self, modules_data):
...
+                tile_last_corner = (
+                    np.array([panel_corner_x, panel_corner_y, 0.])
+                    + np.array([across, 0, 0]) * (cls.frag_fs_pixels + asic_gap_px)
+                    + np.array([0, up, 0]) * (cls.frag_ss_pixels + asic_gap_px)
+                )
+                tile_first_corner = tile_last_corner - tile_size
+
+                tiles.append(GeometryFragment(
+                    corner_pos=tile_first_corner,
+                    ss_vec=np.array([0, 1, 0]),
```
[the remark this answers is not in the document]
takluyver: It's definitely the intention to store the geometry with sub-pixel accuracy so we can pass it to tools like CrystFEL or PyFAI (we haven't yet written code to export those formats for LPD, but it will come).

We could also implement `position_modules_interpolate`, but it's *really* slow - the [example from AGIPD](https://karabo-data.readthedocs.io/en/latest/agipd_geometry.html) takes ~7 seconds to assemble one image. And interpolating blurs the image, because each pixel of the result is affected by several input pixels.

If people want to do accurate analysis of detector data, the only practical way is to bring accurate geometry information into the analysis tools (like PyFAI) and give them un-assembled data. Assembling images is only really suitable for a visual inspection of the data.

karabo_data/geometry2.py
ebadkamil: Nevermind. I think I got it now. It is used later to specify slow and fast scan vectors
`ss_vec=np.array([x_orient, 0, 0]),`

karabo_data/geometry2.py
```diff
@@ -631,6 +737,228 @@ def plot_data(self, modules_data):
...
+                        fs_vec=fs_vec,
+                        ss_pixels=cls.frag_ss_pixels,
+                        fs_pixels=cls.frag_fs_pixels,
+                    ))
+                modules.append(tiles)
+
+        return cls(modules, filename=path)
+
+
+    def inspect(self, frontview=True):
```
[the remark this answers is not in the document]
takluyver: Martin pointed out that this was difficult to use outside a notebook, and I've changed it to use pyplot to create the figure. So it should work more like other matplotlib things now - call `pyplot.show()` if it doesn't show up automatically.

docs/geometry.rst
```diff
@@ -0,0 +1,78 @@
...
+   .. automethod:: position_modules_interpolate
+
+   .. automethod:: inspect
+
+   .. automethod:: compare
+
+LPD-1M
+------
+
+LPD-1M consists of 16 modules of 256×256 pixels each.
```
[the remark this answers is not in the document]
takluyver: Thanks, updating this doc

karabo_data/geometry2.py
takluyver: I'll add a comment to make that clearer - those are sets, so it's checking that between `fs_vec` and `ss_vec` there's one `(0, 1)` and one `(1, 0)`, but it doesn't assert which is which.

At present, LPD always has `fs_vec` in the x dimension, so it doesn't need transposing, and AGIPD always has it in the y dimension, so it does.

takluyver: (But the latter could change if AGIPD was assembled in the 'windmill' layout which I've heard is a possibility)

karabo_data/geometry2.py
takluyver: No, quadrants 1 and 2 are oriented in positive x direction - the first pixel is at the end with lower x, and then it reads across in increasing x direction. Remember that the diagrams all have x increasing right-to-left.

takluyver:
Got a verbal LGTM from Martin, in addition to Ebad's here.

PR #108, opened 2019-02-06
Allow selecting region-of-interest in get_array()

takluyver:
Selecting a 2D region-of-interest would look something like this:

```python
run.get_array('SPB_IRU_INLINEMIC_CAM:daqOutput', 'data.image.pixels',
              roi=by_index[50:100, 20:250])
```

Slices are in the same order as dimensions in the HDF5 file, so probably [y, x] for camera images; this is normal for working with images in arrays. The syntax works for other numbers of dimensions, so you can also select e.g. a 1D region of interest in digitiser data.

Closes gh-100

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -592,8 +592,9 @@ class DataCollection:
         roi: by_index
             The region of interest. This expression selects data in all
             dimensions apart from the first (trains) dimension. If the data
-            holds a 1D array for each train, roi=by_index[:8] would get the
-            first 8 values from every train.
+            holds a 1D array for each entry, roi=by_index[:8] would get the
+            first 8 values from every train. If the data is 2D or more at
+            each entry, selection looks like roi=by_index[:8, 5:10] .
         """
         self._check_field(source, key)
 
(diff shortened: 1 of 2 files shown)


karabo_data/reader.py
```diff
@@ -589,14 +589,28 @@ def get_array(self, source, key, extra_dims=None):
...
+            holds a 1D array for each train, roi=by_index[:8] would get the
+            first 8 values from every train.
         """
         self._check_field(source, key)
+
+        if not isinstance(roi, by_index):
+            raise TypeError("roi parameter must be instance of by_index")
+        else:
+            roi = roi.value
+            if not isinstance(roi, tuple):
```
tmichela: isnt' this always `False`?

tmichela:
Maybe you could just update the doc-string describing how it works with multiple dimensions as you show in this PR description `by_index[a:b, c:d, e:f]`

fangohr:
Nice. 

Can we have an example in the documentation, please? A notebook (or as part of one) would be preferred.

takluyver:
See PR #111 for an example.

PR #109, opened 2019-02-08
Prototype tree view of sources and keys in a data collection

takluyver:
Using code from h5glance. Here's what it currently looks like:

screenshot from 2019-02-08 10-05-52

The copy buttons copy code like:

```python
.get_array('SA3_XTD10_MCP/ADC/1:channel_9.output', 'data.baseline')
```

### DIFF
--- a/karabo_data/htmltree/__init__.py
+++ b/karabo_data/htmltree/__init__.py
@@ -1,0 +1,128 @@
+from htmlgen import (Document, Element, Division, UnorderedList,
+                     ListItem, html_attribute, Link, Script,
+                     )
+from pathlib import Path
+
+_PKGDIR = Path(__file__).parent
+
+class Style(Element):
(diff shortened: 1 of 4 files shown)


tmichela:
Nice!
Do you plan to add as well the console interface?
Would it be easy to add dataset shape and dtype?

tmichela:
Can you print a warning of the method `view_tree()` isn't called in a notebook,
You should also add `htmlgen` to the dependecies

takluyver:
I would like to add some way to see the keys in the terminal, but I don't think the tree view is quite the right thing - I think it's too long without the collapsing nodes. For the Python API, maybe just making `_keys_for_source()` is enough?

It would also be good to have a way to list keys with lsxfel.

tmichela:
Yes, making `_keys_for_source()` public would be very useful I think.

PR #110, opened 2019-02-12
Document what happens when data is missing

takluyver:
@fangohr as we discussed this morning.


tmichela:
LGTM

takluyver:
@fangohr did we want to confirm anything with ITDM? I'm pretty confident this is how missing data behaves.

tmichela:
Can we merge this?

takluyver:
The idea was we'd get ITDM to review it to ensure that I've correctly understood how the DAQ behaves when there's no data available. But I'm moderately confident, so maybe we shouldn't wait for that.

tmichela:
What's written here is correct to my knowledge.
There is some more detail to the question, i.e. some components have a max update rate lower than the 10Hz train rate. Typically all beckhoff devices are updating their value in karabo at 2Hz (by default). which means if a sensor's value change 10 or 100 times per seconds, we'll have update only every 500 ms in the control system. But this is not specific to writing data to files, it is the same state for all the system. I'm not sure if this kind of details is important here.

takluyver:
I've added a brief sentence about that, if you want to take a look.

tmichela:
:+1:

PR #111, opened 2019-02-13
Add example of region-of-interest selection

takluyver:
I also made a couple of object reprs more useful while I was looking at the docs.


fangohr:
LGTM

PR #113, opened 2019-02-18
Major improvements to get_array

takluyver:
This reworks `DataCollection.get_array()` with two major benefits:

1. It works for data with multiple entries per train, which previously it rejected
2. It now selects data by train IDs before loading it, whereas previously it loaded all the data from the files and then threw away the data for unselected trains. This is important because point 1 means that it can be used with fast detector data, so 'load this all into memory' is not a good tactic.

The reworking involves some new internal machinery - the `DataChunk` class and `DataCollection._find_data_chunks`, which I expect to reuse for building virtual datasets.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -740,7 +740,7 @@ class DataCollection:
         else:
             # The translated pattern ends with "\Z" - insert before this
             p = key_re.pattern
-            end_ix = p.rindex('\Z')
+            end_ix = p.rindex(r'\Z')
             ctrl_key_re = re.compile(p[:end_ix] + r'(\.value)?' + p[end_ix:])
 
         matched = {}
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -258,6 +258,34 @@ class FileAccess:
         self._keys_cache[source] = res
         return res
 
+
+class DataChunk:
+    """Reference to a contiguous chunk of data for one or more trains."""
+    def __init__(self, file: FileAccess, source, key, first, train_ids, counts):
+        self.file = file
+        self.source = source
+        self.key = key
+        self.first = first
+        self.train_ids = train_ids
+        self.counts = counts
+
+    @property
+    def slice(self):
+        return slice(self.first, self.first + np.sum(self.counts))
+
+    @property
+    def dataset(self):
+        if self.source in self.file.instrument_sources:
+            group = 'INSTRUMENT'
+        elif self.source in self.file.control_sources:
+            group = 'CONTROL'
+        else:
+            raise SourceNameError(self.source)
+
+        return self.file.file['/{}/{}/{}'.format(
+            group, self.source, self.key.replace('.', '/'))]
+
+
 class DataCollection:
     """An assemblage of data generated at European XFEL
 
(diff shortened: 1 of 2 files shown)


karabo_data/reader.py
```diff
@@ -258,6 +258,34 @@ def add_key(key, value):
...
+        self.counts = counts
+
+    @property
+    def slice(self):
+        return slice(self.first, self.first + np.sum(self.counts))
+
+    @property
+    def dataset(self):
+        if self.source in self.file.instrument_sources:
+            group = 'INSTRUMENT'
```
antarcticrainforest: I know this is pretty unrelated. But how are INSTRUMENT and DETECTOR sources defined? DETECTOR is fast data (pulse resolved) that is 2D per pusle (hence image)? and INSTRUMENT all other fast data? I know that at the moment AGIPD and LPD are the 'only' detectors. But since over the last few month other detectors  have been integrated we might want to add detectors, or at least revisit the definition of what a detector is. I'll try to collect a list of device names of any devices that should fall under the category "DETECTOR" but are INSTRUMENT at the moment, like the Gotthard.

takluyver: There is no DETECTOR group in the files, as far as I know. AGIPD and LPD data are recorded under INSTRUMENT. If this is something you know of that's been changed recently, please point me to an example, because it will probably need prompt work on `karabo_data`.

The top-level groups I'm aware of are INSTRUMENT, CONTROL, RUN, INDEX and METADATA. So far we've ignored RUN because it was duplicating information already in CONTROL (this is set to change as people can be more selective about what is stored in CONTROL). INSTRUMENT and CONTROL are the groups we use for data, and INDEX and METADATA tell us about how to access the data.

antarcticrainforest: Sorry for the confusion, I was referring to the detector_sources property in the reader module, not the HDF5 files. At the moment we have 'only' AGIPD and LPD. I got just curious how you would define a detector? Because as yous said the files do not have a detector group. And based on the AGIPD and LPD example we might want to add more 'detectors' to the detector_sources in karabo_data.

takluyver: With hindsight I probably shouldn't have casually used the term 'detectors' like this. This has come up in #95 as well. Detectors can mean a lot of different things.

My mental model at the time was roughly that each experiment would be centred around one big detector, like AGIPD or LPD, which produces the biggest and most essential data. This is about right for SPB, and I think for FXE as well, but some of the other instruments are doing different kinds of experiments which don't seem to be centred around one huge, specialised detector.

It might be easiest to ultimately drop some 'detector' APIs and move towards more specific functionality, rather than trying to generalise over a set of very different detectors.

takluyver:
@antarcticrainforest were you happy with these changes, or would you like them to have further review (from you or someone else)?

antarcticrainforest:
Nope, I don't have anything to add on my side!!!

takluyver:
Thanks! But to clarify, are you happy that you have reviewed these changes, so I can merge the pull request?

antarcticrainforest:
Sorry, yes that looks all good to me. I don't have any objections -> LGTM

takluyver:
Super, thanks!

karabo_data/reader.py
```diff
@@ -634,39 +662,19 @@ def get_array(self, source, key, extra_dims=None, roi=by_index[...]):
...
-        if source in self.control_sources:
-            data_path = "/CONTROL/{}/{}".format(source, key.replace('.', '/'))
-            for f in self._source_index[source]:
-                slices = (slice(None, len(f.train_ids)),) + roi
-                data = f.file[data_path][slices]
-                if extra_dims is None:
-                    extra_dims = ['dim_%d' % i for i in range(data.ndim - 1)]
-                dims = ['trainId'] + extra_dims
+        for chunk in self._find_data_chunks(source, key):
+            print("Got chunk in", chunk.file)
```
tmichela: is this a debug print that snuck under the radar?

takluyver: It is, good catch! Just removed it: 9de53011d2560090d175d8fb62570bb25eb7acd9

PR #115, opened 2019-02-20
Fix pixel corner positions in distortion array

takluyver:
I previously misread the CrystFEL geometry docs and thought it gave coordinates based on pixel centres. This was incorrect; the coordinates are actually for the first corner of the pixel ('first' in the sense of nearest the start of the ASIC in the order its data is read).

As a result, the distortion array we generate for PyFAI was half a pixel out. This PR corrects that.

cc @ebadkamil


tmichela:
LGTM

takluyver:
Thanks!

PR #117, opened 2019-02-26
Make keys_for_source method public

takluyver:
As discussed on #109, there should be a public Python API to find the keys for a specified source. There is a method to do this, but it was private. This pull request takes the simple approach of making that method public.


fangohr:
Can you add an example in the documentation so that users learn early on that they can use this method to find the available keys? 

Otherwise LGTM.

tmichela:
LGTM

takluyver:
Added an example, merging.

PR #118, opened 2019-03-01
Documentation for accessing AGIPD & LPD data

takluyver:
This adds API documentation and an example for the interfaces added in #95.


antarcticrainforest:
LGTM

Issue #126, opened 2019-03-22
TypeError: intersect1d() got an unexpected keyword argument 'return_indices'


Calpas:
Dear experts,
I run the command (1) which give me (2). But then running (3) give me the error message (4). Do see what is wrong?
Regards

(1)
run = run.select_trains(by_index[:10])
run.instrument_sources

(2)
frozenset({'SPB_DET_AGIPD1M-1/DET/0CH0:xtdf',
           'SPB_DET_AGIPD1M-1/DET/10CH0:xtdf',
           ...
           'SPB_DET_AGIPD1M-1/DET/9CH0:xtdf'})

(3)
data_module0 = run.get_array('SPB_DET_AGIPD1M-1/DET/0CH0:xtdf', 'image.data')

(4)
TypeError                                 Traceback (most recent call last)
<ipython-input-35-b8056678b051> in <module>()
      1 
----> 2 data_module0 = run.get_array('SPB_DET_AGIPD1M-1/DET/0CH0:xtdf', 'image.data')
      3 
      4 #data_module0.shape

/software/anaconda3/5.2/lib/python3.6/site-packages/karabo_data/reader.py in get_array(self, source, key, extra_dims, roi)
    546         seq_arrays = []
    547 
--> 548         for chunk in self._find_data_chunks(source, key):
    549             trainids = self._expand_trainids(chunk.counts, chunk.train_ids)
    550 

/software/anaconda3/5.2/lib/python3.6/site-packages/karabo_data/reader.py in _find_data_chunks(self, source, key)
    791         for file in self._source_index[source]:
    792             trains_of_interest, train_ixs, _ = np.intersect1d(
--> 793                 file.train_ids, self.train_ids, return_indices=True
    794             )
    795             if trains_of_interest.size == 0:

TypeError: intersect1d() got an unexpected keyword argument 'return_indices'

takluyver:
I don't think you're doing anything wrong, but we've used something which requires a relatively new version of numpy (1.15), and it seems that the version on Maxwell is slightly too old (1.14.3).

If you're doing this with SSH to Maxwell, you can use our Anaconda installation by running `module load exfel exfel_anaconda3`. If you're using [max-jhub.desy.de](https://max-jhub.desy.de/), you can run that command in a terminal, and then there should be an 'XFEL' kernel available in the notebook menus.

Calpas:
[...]

PR #127, opened 2019-03-25
Make _find_data_chunks() compatible with numpy 1.14

takluyver:
This should allow karabo_data to work with the main Anaconda installation on Maxwell, which we broke in 0.3 by using the return_indices parameter to `intersect1d()`.

The contiguous_regions function is already used in `hdf5-virtualise`.

Closes gh-126

### DIFF
--- a/karabo_data/read_machinery.py
+++ b/karabo_data/read_machinery.py
@@ -109,6 +109,35 @@ class DataChunk:
         ]
 
 
+# contiguous_regions() by Joe Kington on Stackoverflow
+# https://stackoverflow.com/a/4495197/434217
+# Used here under Stackoverflow's default CC-BY-SA 3.0 license.
+def contiguous_regions(condition):
+    """Finds contiguous True regions of the boolean array "condition". Returns
+    a 2D array where the first column is the start index of the region and the
+    second column is the end index."""
+
+    # Find the indicies of changes in "condition"
+    d = np.diff(condition)
+    idx, = d.nonzero()
+
+    # We need to start things after the change in "condition". Therefore,
+    # we'll shift the index by 1 to the right.
+    idx += 1
+
+    if condition[0]:
+        # If the start of condition is True prepend a 0
+        idx = np.r_[0, idx]
+
+    if condition[-1]:
+        # If the end of condition is True, append the length of the array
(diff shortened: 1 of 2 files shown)


takluyver:
I've confirmed that the tests pass on Maxwell with this branch.

tmichela:
Do we have a test covering `_find_data_chunks`, to make sure the behavior did not change?

takluyver:
`_find_data_chunks()` is called by `get_array()`, so it's covered by the tests there:

https://codecov.io/gh/European-XFEL/karabo_data/pull/127/src/karabo_data/reader.py?before=karabo_data/reader.py#L782...814

I'm a bit puzzled that the last bit doesn't appear to be coverered, though. I thought that `test_run_get_array_empty` would exercise that.

takluyver:
OK, I've understood why the last branch is not covered. I changed the other code in the method a few times while working on it, and I think it's now impossible to hit that branch:

- If no data was recorded for that source, the main branch will still yield chunks referring to trains but with no data.
- If you use `run.select_trains()` to deselect some trains, any file with no matching trains will be dropped from the index completely, so this code won't use it.

I think there might still be a corner case if a file is created with a valid XFEL structure, but no data at all is written to it.

takluyver:
Do we want to leave the uncovered branch in to catch such corner cases, or remove it and wait to see if they come up?

tmichela:
good question.
Such file in a run would probably indicate a problem in the recording... but I can imagine a case where only one data group is failing.

I'd leave it (if ever the case occurs), as long as the file is valid it is probably more correct to return empty datasets...

tmichela:
LGTM

takluyver:
Thanks; here comes 0.4.

tmichela:
should it be `0.3.1` ?

takluyver:
I think 0.4 because we dropped support for Python 3.4, but I think it's sufficiently unlikely that anyone's using that that it doesn't matter much what number we give it.

takluyver:
And I've done it as 0.4 now :wink:

tmichela:
ok no problem, I just try to keep the same pattern on all projects (for my own sanity) as karabo releases uses the 3rd digit for hotfixes.

PR #121, opened 2019-03-11
Test on Python 3.7, drop 3.4

takluyver:
Closes gh-120

PR #128, opened 2019-03-25
Update dependencies for Python 3.5

takluyver:


takluyver:
This is complementary to #121, where we dropped tests on Python 3.4.

tmichela:
LGTM

PR #129, opened 2019-03-25
Add to_distortion_array method for LPD

takluyver:
Generalises the code which already does this for AGIPD.

### DIFF
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -145,7 +145,7 @@ class DetectorGeometryBase:
     pixel_size = 0.0
     frag_ss_pixels = 0
     frag_fs_pixels = 0
-    expected_data_shape = ()
+    expected_data_shape = (0, 0, 0)
 
     def __init__(self, modules, filename='No file'):
         # List of lists (1 per module) of fragments (1 per tile)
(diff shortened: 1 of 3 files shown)


karabo_data/geometry2.py
```diff
@@ -976,6 +997,38 @@ def split_tiles(module_data):
...
+    def _distortion_array_slice(cls, m, t):
+        # Which part of the array is this tile?
+        # The distortion array for LPD is 4096 x 256, with modules stacked
+        # along their slow-scan axis, Q1M1 (m=0) to Q4M4 (m=15)
+        module_offset = m * 256
+        if t < 8:  # First half of module (0 <= t <= 7)
+            fs_slice = slice(0, 128)
+            tiles_up = 7 - t
+        else:      # Second half of module (8 <= t <= 15)
+            fs_slice = slice(128, 256)
```
takluyver: Pre-emptively, because I'm sure @tmichela will ask: I tried writing this method with `cls.frag_ss_pixels` etc. instead of hardcoded numbers. But I found it was much harder to read when I did that, so I went back to hardcoded numbers. This is on the specific LPD class anyway, so the numbers can't really change.

tmichela:
Is there any easy way to test with pyFAI if the generated file is compatible and behave as we expect?

 code LGTM, maybe @antarcticrainforest can also have a look here.

takluyver:
I suspect it's not that hard for someone who knows pyFAI to get some LPD data and try applying the distortion array to it, but as I don't know pyFAI I'm hoping one of the people in the group who does can try that. I suspect Ebad is the best person to try it, but I saw that he's away for some time.

tmichela:
I'm happy to merge this if it is blocking other development.
Maybe @ebadkamil  or @antarcticrainforest  can review afterwards

antarcticrainforest:
Yes, I also think that we can merge this.

takluyver:
Thanks, I'll merge this then. Any further review is welcome!

PR #132, opened 2019-03-27
Write files with virtual datasets

takluyver:
Add a `DataCollection.write_virtual()` method to write a file containing virtual datasets for all selected sources. This provides a convenient way to join up the data from several sequence files.

This does *not* attempt to do what `hdf5-virtualise` does - that's a 2D combination along sequence files and across detector modules, which is more complex than the 1D combination here. I do plan to integrate that into karabo_data, but not in this PR.

I've added a test, but I'll try it on real data this afternoon.

### DIFF
--- a/docs/reading_files.rst
+++ b/docs/reading_files.rst
@@ -58,6 +58,8 @@ including the main detectors, record data more frequently.
 
    .. automethod:: write
 
+   .. automethod:: write_virtual
+
 Missing data
 ------------
 
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -937,6 +937,16 @@ class DataCollection:
         FileWriter(filename, self).write()
 
     def write_virtual(self, filename):
+        """Write an HDF5 file with virtual datasets for the selected data.
+
+        This doesn't copy the data, but each virtual dataset provides a view of
+        data spanning multiple sequence files, which can be accessed as if it
+        had been copied into one big file.
+
+        Creating and reading virtual datasets requires HDF5 version 1.10.
+
+        The target filename will be overwritten if it already exists.
+        """
         from .writer import VirtualFileWriter
         VirtualFileWriter(filename, self).write()
 
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -936,6 +936,10 @@ class DataCollection:
         from .writer import FileWriter
         FileWriter(filename, self).write()
 
+    def write_virtual(self, filename):
+        from .writer import VirtualFileWriter
+        VirtualFileWriter(filename, self).write()
+
 
 class TrainIterator:
     """Iterate over trains in a collection of data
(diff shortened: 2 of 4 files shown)


tmichela:
Nice!

nitpicking: add maybe a small docstring to `DataCollection.write_virtual`
LGTM

takluyver:
Added a docstring.

I've also now tested it on the code, and after some changes I was successfully able to build a virtual file for a run, and the virtual file passes `karabo-data-validate`.

tmichela:
I did not have time to test yet, but: if you merge 2 `DataConnection` with a large number of train gap in between how does it work? Does it:
- Fill all trainIds and put count to 0 for all?
- leave a gap between missing trainIds?

takluyver:
It should leave missing train IDs out completely, so it will store e.g. `[1, 2, 3, 20, 21, 22]`. I should try to test this, though.

tmichela:
Tried it and all datasets are full of zeroes. Am I doing something wrong?

```python
path = '/gpfs/exfel/exp/XMPL/201750/p700000/raw/r0006'
run = RunDirectory(path)
run.write_virtual('full.h5')
f = h5py.File('full.h5')
ds = f['INSTRUMENT/SPB_DET_AGIPD1M-1/DET/9CH0:xtdf/image/trainId'][:]
 
array([[0],
       [0],
       [0],
       ...,
       [0],
       [0],
       [0]], dtype=uint64)
```

tmichela:
Ah, you should also update the minimal version for `h5py`

takluyver:
Did you not get an error from that about `VirtualLayout` being missing? Maybe we should just do an explicit check for that before it creates the file.

takluyver:
(Rebased to pick up the new name for a test fixture, and force-pushed)

tmichela:
Yes, I got the error. After updating the environment the file gets written, though all data sets are empty (shapes seem correct, but all filled with zeros)

takluyver:
Can you do an access check on it, using the script from hdf5-virtualise?

```shell
module load exfel hdf5-virtualise
hdf5-vds-access-check full.h5
```

This check should work on any file containing virtual datasets. It might be slow if the file contains a large number of virtual datasets, though.

tmichela:
```bash
~/projects/karabo_data% hdf5-vds-access-check full.h5
Checking virtual dataset: INSTRUMENT/SPB_DET_AGIPD1M-1/DET/0CH0:xtdf/detector/data
  1/1 sources accessible

[...]

Checking virtual dataset: INSTRUMENT/SPB_DET_AGIPD1M-1/DET/9CH0:xtdf/trailer/status
  1/1 sources accessible

All virtual data sources accessible
```

tmichela:
Okay then, PR is fine on my side

tmichela:
This is a very annoying behavior :(

takluyver:
It is

PR #133, opened 2019-03-27
Mock agipd run

antarcticrainforest:
This includes a mock run for SPB. Also I have added a raw flag to create either raw or calibrated mock data (fxe/spb).

### DIFF
--- a/karabo_data/tests/conftest.py
+++ b/karabo_data/tests/conftest.py
@@ -50,19 +50,13 @@ def mock_spb_control_data_badname():
 
 
 @pytest.fixture(scope='module')
-def mock_fxe_run():
+def mock_fxe_raw_run():
     with TemporaryDirectory() as td:
         make_examples.make_fxe_run(td)
         yield td
 
 @pytest.fixture(scope='module')
-def mock_spb_run():
-    with TemporaryDirectory() as td:
-        make_examples.make_spb_run(td)
-        yield td
-
-@pytest.fixture(scope='module')
-def mock_spb_run():
+def mock_spb_proc_run():
     with TemporaryDirectory() as td:
         make_examples.make_spb_run(td, raw=False)
         yield td
--- a/karabo_data/tests/conftest.py
+++ b/karabo_data/tests/conftest.py
@@ -55,6 +55,11 @@ def mock_fxe_run():
         make_examples.make_fxe_run(td)
         yield td
 
+@pytest.fixture(scop='module')
+def mock_spb_run():
+    with TemporaryDirectory() as td:
+        make_examples.make_spb_run(td)
+        yield td
 
 @pytest.fixture(scope='module')
 def empty_h5_file():
--- a/karabo_data/tests/conftest.py
+++ b/karabo_data/tests/conftest.py
@@ -55,10 +55,16 @@ def mock_fxe_run():
         make_examples.make_fxe_run(td)
         yield td
 
-@pytest.fixture(scop='module')
+@pytest.fixture(scope='module')
 def mock_spb_run():
     with TemporaryDirectory() as td:
         make_examples.make_spb_run(td)
+        yield td
+
+@pytest.fixture(scope='module')
+def mock_spb_run():
+    with TemporaryDirectory() as td:
+        make_examples.make_spb_run(td, raw=False)
         yield td
 
 @pytest.fixture(scope='module')
--- a/karabo_data/tests/make_examples.py
+++ b/karabo_data/tests/make_examples.py
@@ -215,25 +215,53 @@ def make_lpd_file(path):
         LPDModule('FXE_DET_LPD1M-1/DET/0CH0', frames_per_train=128)
     ], ntrains=480, chunksize=32)
 
-def make_fxe_run(dir_path):
+def make_fxe_run(dir_path, raw=True):
+    prefix = 'RAW' if raw else 'CORR'
     for modno in range(16):
-        path = osp.join(dir_path, 'RAW-R0450-LPD{:0>2}-S00000.h5'.format(modno))
+        path = osp.join(dir_path,
+                        '{}-R0450-LPD{:0>2}-S00000.h5'.format(prefix, modno))
         write_file(path, [
             LPDModule('FXE_DET_LPD1M-1/DET/{}CH0'.format(modno), frames_per_train=128)
         ], ntrains=480, chunksize=32)
-
+    if not raw:
+        return
     write_file(osp.join(dir_path, 'RAW-R0450-DA01-S00000.h5'), [
         XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
         XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
         GECCamera('FXE_XAD_GEC/CAM/CAMERA'),
         GECCamera('FXE_XAD_GEC/CAM/CAMERA_NODATA', nsamples=0),
     ], ntrains=400, chunksize=200)
-    write_file(osp.join(dir_path, 'RAW-R0450-DA01-S00001.h5'), [
+    write_file(osp.join(dir_path, '{}-R0450-DA01-S00001.h5'.format(prefix)), [
         XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
         XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
         GECCamera('FXE_XAD_GEC/CAM/CAMERA'),
         GECCamera('FXE_XAD_GEC/CAM/CAMERA_NODATA', nsamples=0),
     ], ntrains=80, firsttrain=10400, chunksize=200)
+
+def make_spb_run(dir_path, raw=True):
+    prefix = 'RAW' if raw else 'CORR'
+    for modno in range(16):
+        path = osp.join(dir_path,
+                        '{}-R0238-AGIPD{:0>2}-S00000.h5'.format(prefix, modno))
+        write_file(path, [
+            AGIPDModule('SPB_DET_AGIPD1M-1/DET/{}CH0'.format(modno),
+                         frames_per_train=64)
+            ], ntrains=32, chunksize=32)
+    if not raw:
+        return
+    write_file(osp.join(dir_path, 
+                        '{}-R0238-DA01-S00000.h5'.format(prefix)),
+                [
+                    XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
+                    XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                ], ntrains=32, chunksize=32)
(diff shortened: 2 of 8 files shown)


karabo_data/tests/make_examples.py
```diff
@@ -243,4 +272,6 @@ def make_fxe_run(dir_path):
     make_lpd_file('lpd_example.h5')
     os.makedirs('fxe_example_run', exist_ok=True)
     make_fxe_run('fxe_example_run')
+    os.mkdirs('spb_example_run', exist_ok=True)
```
takluyver: `mkdirs` -> `makedirs`

karabo_data/tests/conftest.py
```diff
@@ -55,6 +55,17 @@ def mock_fxe_run():
...
         yield td
 
+@pytest.fixture(scope='module')
+def mock_spb_run():
+    with TemporaryDirectory() as td:
+        make_examples.make_spb_run(td)
+        yield td
+
+@pytest.fixture(scope='module')
+def mock_spb_run():
```
tmichela: You have 2 fixtures with the same name

karabo_data/tests/mockdata/detectors.py
```diff
@@ -17,19 +17,26 @@ class DetectorModule:
...
+        if self.raw:
+            data_type = 'u2'
+        else:
+            data_type = 'f4'
         return [
             ('cellId', 'u2', (1,)),
-            ('data', 'u2', self.image_dims),
+            ('data', data_type, self.image_dims),
             ('length', 'u4', (1,)),
             ('status', 'u2', (1,)),
```
takluyver: calibrated data also has mask and gain datasets, IIRC

karabo_data/tests/conftest.py
takluyver: I think these should have different names, otherwise only one of them can be used. Maybe `mock_spb_run_raw` and `mock_spb_run_proc`?

karabo_data/tests/make_examples.py
```diff
@@ -215,26 +215,55 @@ def make_lpd_file(path):
...
+                [
+                    XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
+                    XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                ], ntrains=32, chunksize=32)
+    write_file(osp.join(dir_path,
+                        '{}-R0238-DA01-S00001.h5'.format(prefix)),
+                [
+                    XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
+                    XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                ], ntrains=32, firsttrain=10032, chunksize=32)
```
takluyver: You've got 32 trains for the AGIPD data, but a total of 64 for the other sources, between these two sequence files.

karabo_data/tests/mockdata/detectors.py
```diff
@@ -17,19 +17,26 @@ class DetectorModule:
         'trailer',
     ]
 
-    def __init__(self, device_id, frames_per_train=64):
+    def __init__(self, device_id, frames_per_train=64, raw=True):
         self.device_id = device_id
         self.frames_per_train = frames_per_train
+        if not raw:
+            self.image_dims = self.image_dims[1:]
```
takluyver: Maybe add a comment about this - I think I know why it's needed, but it's a bit mysterious.

karabo_data/tests/conftest.py
antarcticrainforest: Ups, that was not intentional, I'll delete one. Since mock runs can be raw and proc I figured that spb could be a test for prc and fxe for raw?

karabo_data/tests/mockdata/detectors.py
antarcticrainforest: Yes good point! I'll add them

karabo_data/tests/conftest.py
antarcticrainforest: I thought one might only check one run (fxe/spb) for proc and the other (spb/fxe) for raw. Having to two fixtures is of course nonsense. I'll delete one

takluyver: Yup, that's also fine to have a calibrated SPB run and a raw FXE run for testing. I'd still suggest naming the new one with `proc` or `calib` in the name to make that clear.

karabo_data/tests/make_examples.py
antarcticrainforest: Ahh gotcha! Now that makes sense to me! I'll change the AGIPD data to 64 modules as well.

karabo_data/tests/mockdata/detectors.py
antarcticrainforest: Okidoke

karabo_data/tests/conftest.py
antarcticrainforest: I've changed the fixture names for mock_fxe/spb_run and added raw/proc to the method name depending on what they kind of data they represent. But I think that will clash with the other PR I've just seen. You want me to rename mock_fxe_raw_run back to mock_fxe_run? If done then I also would like to rename mock_spb_proc_run back to mock_spb_run for consitency.

karabo_data/tests/mockdata/detectors.py
```diff
@@ -17,22 +17,37 @@ class DetectorModule:
         'trailer',
     ]
 
-    def __init__(self, device_id, frames_per_train=64):
+    def __init__(self, device_id, frames_per_train=64, raw=True):
         self.device_id = device_id
         self.frames_per_train = frames_per_train
+        if not raw:
+            # Drop the first dim (gain) if not raw (proc data)
```
takluyver: This is a bit of a 'repeat the code in English' comment ;-) How about something like:

```python
# Raw data has an extra dimension, used in AGIPD to separate data and gain.
# This dimension is removed by the calibration process.
```

karabo_data/tests/mockdata/detectors.py
```diff
@@ -17,22 +17,37 @@ class DetectorModule:
...
+                ('length', 'u4', (1,)),
+                ('status', 'u2', (1,)),
+            ]
+
+        else:
+            return [
+                ('cellId', 'u2', (1,)),
+                ('data', 'f4', self.image_dims),
+                ('mask', 'u2', self.image_dims),
+                ('gain', 'u2', self.image_dims),
```
takluyver: In an example I'm looking at (`/gpfs/exfel/exp/SPB/201701/p002012/raw/r0243`), mask and gain are uint8, which is `'u1'` in this system (1 byte). Do you have an example where they're 2 bytes?

karabo_data/tests/mockdata/detectors.py
```diff
@@ -17,22 +17,37 @@ class DetectorModule:
...
+            return [
+                ('cellId', 'u2', (1,)),
+                ('data', 'u2', self.image_dims),
+                ('length', 'u4', (1,)),
+                ('status', 'u2', (1,)),
+            ]
+
+        else:
+            return [
+                ('cellId', 'u2', (1,)),
```
takluyver: Also in the example I'm looking at, `cellId` drops the extra dimension denoted by `(1,)` here.

karabo_data/tests/test_reader_mockdata.py
```diff
@@ -105,23 +105,39 @@ def test_iterate_trains_require_all(mock_sa3_control_data):
...
+    device = 'SPB_DET_AGIPD1M-1/DET/15CH0:xtdf'
+    assert tid == 10000
+    assert 'image.data' in data[device]
+    assert 'f4' == data[device]['image.data'].dtype
+    for prop in ('image.gain', 'image.mask'):
+        assert prop in data[device]
+        assert 'u2' == data[device][prop].dtype
+
+
+
```
takluyver: Excess whitespace

karabo_data/tests/test_reader_mockdata.py
```diff
@@ -105,23 +105,39 @@ def test_iterate_trains_require_all(mock_sa3_control_data):
...
 
+def test_read_spb_proc_run(mock_spb_proc_run):
+    run = RunDirectory(mock_spb_proc_run) #Test for calib data
+    assert len(run.files) == 16 # only 16 detector modules for calib data
+    assert run.train_ids == list(range(10000, 10064)) #64 trains
+    tid, data = next(run.trains())
+    device = 'SPB_DET_AGIPD1M-1/DET/15CH0:xtdf'
+    assert tid == 10000
+    assert 'image.data' in data[device]
+    assert 'f4' == data[device]['image.data'].dtype
```
takluyver: Did you run this? I'd be surprised if comparing a dtype object against a string like this works.

karabo_data/tests/conftest.py
takluyver: No, that's OK, the new name makes sense. We'll just have to reconcile the difference after we merge one of the PRs.

karabo_data/tests/test_reader_mockdata.py
antarcticrainforest: Yes, I did and it did not complain - I mean the travis checks were also successful, right? But to be honest I was not thinking about it when wrote this, now that you mention it I am surprised that this is working!

karabo_data/tests/mockdata/detectors.py
antarcticrainforest: I just checked one of the more recent (201930) calibrated runs and the mask is u4 whereas the gain is u1. I'll fix that

karabo_data/tests/mockdata/detectors.py
antarcticrainforest: Ok I think I did not understand the shapes in the first place ... ('foo', 'bar', (1, )) will create an array of shape (ntrains x npulses, 1) and ('foo', 'bar', ()) will create an array of shape (ntrains x npulses)?

karabo_data/tests/mockdata/detectors.py
antarcticrainforest: Okidoke

karabo_data/tests/mockdata/detectors.py
takluyver: Yup, that's right. The tuple gives any extra dimensions for the dataset after the first (trains x pulses) dimension.

karabo_data/tests/test_reader_mockdata.py
takluyver: Oh yes, it did work. I had tried `'u8' == np.uint64`, which doesn't work, but the relevant thing is `'u8' == np.dtype(np.uint64)`, which does.

takluyver:
LGTM, assuming tests pass again. Travis is [having difficulties](https://www.traviscistatus.com/) at the moment.

PR #135, opened 2019-03-28
Add basler camera to mock data

antarcticrainforest:
Yet another, short PR, to add the Basler camera to mock data.

### DIFF
--- a/karabo_data/tests/make_examples.py
+++ b/karabo_data/tests/make_examples.py
@@ -3,6 +3,20 @@ import os
 import os.path as osp
 import numpy as np
 
+from .mockdata import write_file
+from .mockdata.xgm import XGM
+from .mockdata.gec_camera import GECCamera
+from .mockdata.basler_camera import BaslerCamera as BaslerCam
+from .mockdata.adc import ADC
+from .mockdata.uvlamp import UVLamp
+from .mockdata.motor import Motor
+from .mockdata.tsens import TemperatureSensor
+from .mockdata.imgfel import IMGFELCamera, IMGFELMotor
+from .mockdata.gauge import Gauge
+from .mockdata.dctrl import DCtrl
+from .mockdata.mpod import MPOD
+from .mockdata.detectors import AGIPDModule, LPDModule
+
 vlen_bytes = h5py.special_dtype(vlen=bytes)
 
 def make_metadata(h5file, data_sources, chunksize=16):
@@ -135,21 +149,6 @@ def make_agipd_example_file(path):
     f.create_dataset('INSTRUMENT/SPB_DET_AGIPD1M-1/DET/7CH0:xtdf/trailer/status',
                         (256,), 'u8', maxshape=(None,))  # Empty in example
 
-
-from .mockdata import write_file
-from .mockdata.xgm import XGM
-from .mockdata.gec_camera import GECCamera
-from .mockdata.basler_camera import BaslerCamera
-from .mockdata.adc import ADC
-from .mockdata.uvlamp import UVLamp
-from .mockdata.motor import Motor
-from .mockdata.tsens import TemperatureSensor
-from .mockdata.imgfel import IMGFELCamera, IMGFELMotor
-from .mockdata.gauge import Gauge
-from .mockdata.dctrl import DCtrl
-from .mockdata.mpod import MPOD
-from .mockdata.detectors import AGIPDModule, LPDModule
-
 def make_fxe_da_file(path):
     """Make the structure of a file with non-detector data from the FXE experiment
 
@@ -202,7 +201,7 @@ def make_sa3_da_file(path):
 def make_data_file_bad_device_name(path):
     """Not all devices have the Karabo standard A/B/C naming convention"""
     write_file(path, [
-        BaslerCamera('SPB_IRU_SIDEMIC_CAM', sensor_size=(1000, 1000))
+        BaslerCam('SPB_IRU_SIDEMIC_CAM', sensor_size=(1000, 1000))
     ], ntrains=500, chunksize=50)
 
 def make_agipd_file(path):
@@ -250,22 +249,17 @@ def make_spb_run(dir_path, raw=True, sensor_size=(1024, 768)):
             ], ntrains=64, chunksize=32)
     if not raw:
         return
-    write_file(osp.join(dir_path,
-                        '{}-R0238-DA01-S00000.h5'.format(prefix)),
-                [
-                    XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
-                    XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
-                    BaslerCamera('SPB_IRU_CAM/CAM/SIDEMIC',
-                        sensor_size=sensor_size), ], ntrains=32, chunksize=32)
-    write_file(osp.join(dir_path,
-                        '{}-R0238-DA01-S00001.h5'.format(prefix)),
-                [
-                    XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
-                    XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
-                    BaslerCamera('SPB_IRU_CAM/CAM/SIDEMIC',
-                        sensor_size=sensor_size), ], ntrains=32,
-                    firsttrain=10032, chunksize=32)
+    write_file(osp.join(dir_path, '{}-R0238-DA01-S00000.h5'.format(prefix)),
+               [ XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
+                 XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                 BaslerCam('SPB_IRU_CAM/CAM/SIDEMIC', sensor_size=sensor_size)
+               ], ntrains=32, chunksize=32)
 
+    write_file(osp.join(dir_path, '{}-R0238-DA01-S00001.h5'.format(prefix)),
+               [ XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
+                 XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                 BaslerCam('SPB_IRU_CAM/CAM/SIDEMIC', sensor_size=sensor_size)
+               ], ntrains=32, firsttrain=10032, chunksize=32)
 
 if __name__ == '__main__':
     make_agipd_example_file('agipd_example.h5')
--- a/karabo_data/tests/make_examples.py
+++ b/karabo_data/tests/make_examples.py
@@ -202,7 +202,7 @@ def make_sa3_da_file(path):
 def make_data_file_bad_device_name(path):
     """Not all devices have the Karabo standard A/B/C naming convention"""
     write_file(path, [
-        BaslerCamera('SPB_IRU_SIDEMIC_CAM', px_size=(1000, 1000))
+        BaslerCamera('SPB_IRU_SIDEMIC_CAM', sensor_size=(1000, 1000))
     ], ntrains=500, chunksize=50)
 
 def make_agipd_file(path):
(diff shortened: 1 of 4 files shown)


karabo_data/tests/make_examples.py
```diff
@@ -140,6 +140,7 @@ def make_train_ids(path):
 from .mockdata.xgm import XGM
 from .mockdata.gec_camera import GECCamera
 from .mockdata.sidemic_camera import SidemicCamera
+from .mockdata.basler_camera import BaslerCamera
```
tmichela: The component called `SidemicCamera` is actually a basler camera. You could maybe then remove the other one. And since there are various Basler reference with difference resolution it would be also nice if this was configurable

karabo_data/tests/make_examples.py
```diff
@@ -255,12 +256,14 @@ def make_spb_run(dir_path, raw=True):
                 [
                     XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
                     XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                    BaslerCamera('SPB_IRU_SIDEMIC_CAM'),
```
tmichela: the naming convention is not respected in the camera you saw, but I think it would be better if you make up a name that respect it here. e.g. `SPB_IRU_CAM/CAM/SIDEMIC`

karabo_data/tests/mockdata/basler_camera.py
```diff
@@ -0,0 +1,35 @@
...
+    """
+
+    output_channels = ('daqQutput/data',)
+
+    instrument_keys = [
+        ('image/bitsPerPixel', 'i4', ()),
+        ('image/dimTypes', 'i4', (2,)),
+        ('image/dims', 'u8', (2,)),
+        ('image/encoding', 'i4', ()),
+        ('image/pixels', 'u2', (1029, 1228)), # Odd shape
```
tmichela: did you see that shape in a file? reason is probable due to ROI settings. Either find the native resolution, or set the ROI information accordingly.

karabo_data/tests/make_examples.py
takluyver: Sorry about the misnaming there.

tmichela: No worries, I probably did not know myself back then :)

tmichela: But also that raises again the point that it is impossible to know what hardware is hidden behind a name in files unless one has expert knowledge of the instrument setup...
I believe that the karabo device class of each device monitored will be stored in the new file format (under the `RUN` group) which might be enough most of the time

takluyver: Oh yes, that would be good. If it is, we should ensure there's a way to get it in karabo_data.

karabo_data/tests/mockdata/basler_camera.py
antarcticrainforest: I guess it's something with the binning. Anyhoo since there are apparently different types of cams out there I added the pixel size a a keyword.

karabo_data/tests/make_examples.py
antarcticrainforest: For now, I've replaced the ```SidemicCamera``` with the ```BaslerCamera``` class

karabo_data/tests/make_examples.py
antarcticrainforest: @tmichela you have a reference dataset?

antarcticrainforest:
Any suggestions?

karabo_data/tests/make_examples.py
```diff
@@ -239,7 +239,7 @@ def make_fxe_run(dir_path, raw=True):
         GECCamera('FXE_XAD_GEC/CAM/CAMERA_NODATA', nsamples=0),
     ], ntrains=80, firsttrain=10400, chunksize=200)
 
-def make_spb_run(dir_path, raw=True):
+def make_spb_run(dir_path, raw=True, px_size=(1024, 768)):
```
tmichela: `sensor_size` maybe?

karabo_data/tests/make_examples.py
```diff
@@ -255,12 +255,14 @@ def make_spb_run(dir_path, raw=True):
                 [
                     XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
                     XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                    BaslerCamera('SPB_IRU_SIDEMIC_CAM', px_size=px_size),
```
tmichela: Please use a name that follow the naming convention as I mentioned before

karabo_data/tests/make_examples.py
```diff
@@ -255,12 +255,14 @@ def make_spb_run(dir_path, raw=True):
...
                 ], ntrains=32, chunksize=32)
     write_file(osp.join(dir_path,
                         '{}-R0238-DA01-S00001.h5'.format(prefix)),
                 [
                     XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
                     XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
+                    BaslerCamera('SPB_IRU_SIDEMIC_CAM', px_size=px_size),
```
tmichela: dito

karabo_data/tests/mockdata/basler_camera.py
```diff
@@ -0,0 +1,38 @@
...
+    """
+    Basler Camera device
+    Based on example /gpfs/exfel/exp/SPB/201930/p900061/raw/r0055/RAW-R0055-DA01-S00000.h5
+    """
+
+    def __init__(self, device_id, nsamples=None, px_size=None):
+        """Create a dummy basler device that inherits from Device Base"""
+        self.px_size = px_size or (2058, 2456)
+        super(BaslerCamera, self).__init__(device_id, nsamples=nsamples)
+        self.output_channels = ('daqOutput/data',)
```
tmichela: Can you add back the comment from @takluyver 
```python
# Technically, only the part before the / is the output channel.
# But there is a structure associated with the part one level after that,
# and we don't know what else to call it.
```

karabo_data/tests/mockdata/basler_camera.py
```diff
@@ -0,0 +1,38 @@
...
+        ]
+
+    def write_instrument(self, f):
+        super().write_instrument(f)
+
+        # Add fixed metadata
+        for channel in self.output_channels:
+            image_grp = 'INSTRUMENT/{}:{}/image/'.format(self.device_id, channel)
+            f[op_join(image_grp, 'bitsPerPixel')][:self.nsamples] = 16
+            f[op_join(image_grp, 'dims')][:self.nsamples] = self.px_size
```
tmichela: This is the image size according to ROI information, right?

karabo_data/tests/make_examples.py
antarcticrainforest: Instead of px_size? Ok

karabo_data/tests/make_examples.py
antarcticrainforest: Ah yes, will do

karabo_data/tests/mockdata/basler_camera.py
antarcticrainforest: Yes I guess so, I'll double with Gabriele tmr

karabo_data/tests/mockdata/basler_camera.py
antarcticrainforest: Okidoke

karabo_data/tests/mockdata/basler_camera.py
antarcticrainforest: After thinking about it I am not sure about this anymore, it might vary from cam to cam??? I have to admit I have a very profound  cluelessness on the topic!

takluyver:
Do you want to get rid of the misnamed 'sidemic_camera' module while you're looking at this?

karabo_data/tests/test_reader_mockdata.py
```diff
@@ -26,9 +26,10 @@ def test_get_train_bad_device_name(mock_spb_control_data_badname):
     with H5File(mock_spb_control_data_badname) as f:
         train_id, data = f.train_from_id(10004)
         assert train_id == 10004
-        assert 'SPB_IRU_SIDEMIC_CAM:daqOutput' in data
-        assert 'data.image.dims' in data['SPB_IRU_SIDEMIC_CAM:daqOutput']
-        dims = data['SPB_IRU_SIDEMIC_CAM:daqOutput']['data.image.dims']
+        device = 'SPB_IRU_SIDEMIC_CAM:daqOutput'
+        assert device in data.keys()
+        assert 'data.image.dims' in data[device].keys()
```
takluyver: NB you don't need the `.keys()` here - `key in dict` works.

karabo_data/tests/mockdata/basler_camera.py
```diff
@@ -0,0 +1,41 @@
...
+            ('image/flipY', 'u1', ())
+        ]
+
+    def write_instrument(self, f):
+        super().write_instrument(f)
+
+        # Add fixed metadata
+        for channel in self.output_channels:
+            image_grp = 'INSTRUMENT/{}:{}/image/'.format(self.device_id, channel)
+            f[op_join(image_grp, 'bitsPerPixel')][:self.nsamples] = 16
```
takluyver: I wouldn't actually use `os.path` for object paths inside the HDF5 file - these paths always use `/`, regardless of platform. 

I think you can do `f[image_grp]['bitsPerPixel']` if you don't want to join strings.

karabo_data/tests/mockdata/basler_camera.py
takluyver: I believe that when we investigated at some point, `roiOffsets` and `dims` together defined the ROI. But don't rely on me for this.

@tmichela are you happy with using `sensor_size` for this?

tmichela: Oh right, `roiOffsets` define the top left corner and `dims` the x and y dimensions of the stored image.
Yes I'm happy with this

karabo_data/tests/mockdata/basler_camera.py
antarcticrainforest: Ohh....well caught!!!

karabo_data/tests/mockdata/basler_camera.py
```diff
@@ -0,0 +1,41 @@
+"""Script that creates a mock-run for the basler camera"""
+from os.path import join as op_join
```
takluyver: This import is unused now

karabo_data/tests/make_examples.py
```diff
@@ -255,13 +255,16 @@ def make_spb_run(dir_path, raw=True):
...
+                    BaslerCamera('SPB_IRU_CAM/CAM/SIDEMIC',
+                        sensor_size=sensor_size), ], ntrains=32, chunksize=32)
     write_file(osp.join(dir_path,
                         '{}-R0238-DA01-S00001.h5'.format(prefix)),
                 [
                     XGM('SA1_XTD2_XGM/DOOCS/MAIN'),
                     XGM('SPB_XTD9_XGM/DOOCS/MAIN'),
-                ], ntrains=32, firsttrain=10032, chunksize=32)
+                    BaslerCamera('SPB_IRU_CAM/CAM/SIDEMIC',
+                        sensor_size=sensor_size), ], ntrains=32,
```
takluyver: The position of the `]` here makes it hard to see where the list ends - can you tweak the formatting a bit?

karabo_data/tests/test_reader_mockdata.py
```diff
@@ -125,6 +126,16 @@ def test_read_spb_proc_run(mock_spb_proc_run):
     assert 'f4' == data[device]['image.data'].dtype
     run.info() # Smoke test
 
+def test_iterate_spb_raw_run(mock_spb_raw_run):
+    run = RunDirectory(mock_spb_raw_run)
+    trains_iter = run.trains()
+    tid, data = next(trains_iter)
+    assert tid == 10000
+    device = 'SPB_IRU_CAM/CAM/SIDEMIC:daqOutput'
+    assert tid == 10000
```
takluyver: You've asserted this twice

takluyver:
Thanks!

PR #138, opened 2019-04-02
Deprecate selection parameters in stack_detector_data

takluyver:
I want to get rid of these both to simplify the API and to encourage people towards selecting the data of interest before reading it, rather than reading everything and then discarding the unwanted parts. Depending on what data you're looking at, reading data from every source might be very slow.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -20,6 +20,7 @@ import os.path as osp
 import pandas as pd
 import re
 import sys
+from warnings import warn
 import xarray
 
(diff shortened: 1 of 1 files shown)


fangohr:
Great idea.

Can you review the documentation to get rid of occurrences of `xcept=` and `only=` and replace with the appropriate `select` statements instead?

takluyver:
There doesn't appear to be any use of these parameters in the documentation.

tmichela:
LGTM

Issue #139, opened 2019-04-02
unposition all modules


hjkirkwood:
Dear experts,

I have created a mask based on a pixels distance from the center of the assembled agipd image and would like to un-assemble it and go back to the stacked data shape of (16,512,128) based on the geometry file. 
Could that be implemented? I had a look at the code but got a bit lost.

The idea is that I could apply the following and then undo the assemble operation:

```
geom = AGIPD_1MGeometry.from_crystfel_geom(geom_file)

assem_im, center = geom.position_modules_fast(np.ones((16,512,128)))

xg, yg = np.mgrid[0:assem_im.shape[0], 0:assem_im.shape[1]]
radius = np.hypot(xg - center[0],yg - center[1])

rmin = 450
rmax = 550
radius[( (radius<rmin) | (radius>rmax) )]=0
radial_mask = (radius>0)
```

takluyver:
This ties in with something we were discussing yesterday - I think the geometry classes should have methods to convert between locations in data space (module number, slow scan, fast scan) and real space (x, y, z).

hjkirkwood:
Yes this would also be very useful too, I assume you mean to convert from data space to real space and vice versa also. I'd be very interested to have this functionality! Thanks

takluyver:
I might wait until #129 is merged to tackle this, because I think it might be able to share some of the code.

takluyver:
@hjkirkwood I think I've worked out an efficient way to do this kind of thing: #142 would add a `geom.get_pixel_positions()` method giving you an array of (x, y, z) positions for each pixel, in the same shape as one frame of data. So writing a radial mask should be something like this (not tested):

```python
pixelpos = geom.get_pixel_positions()  # shape: (16, 512, 128, 3)

radius = np.sqrt(np.sum(pixelpos ** 2, axis=-1))  # shape: (16, 512, 128)
radius_mask = (radius > 450) & (radius < 550)  # shape: (16, 512, 128)
```

(This is actually calculating the radius in 3D, i.e. it masks all points between two spheres. If the detector is flat it should be the same; or you can discard the Z coordinates to do it in 2D)

Let me know if the API works for you and if there are any changes you want. Once we've agreed how it should work, I'll add docs and tests.

antarcticrainforest:
What happens here if you wanted to do more complicated operations or masking certain ROI's in the 2D plane?

takluyver:
[...]

PR #142, opened 2019-04-04
Add geometry methods to get physical locations of data coordinates

takluyver:
Discussion on #139


karabo_data/geometry2.py
```diff
@@ -335,6 +335,68 @@ def to_distortion_array(self):
...
+        If centre=True, the coordinates are calculated for the centre of each
+        pixel. If not, the coordinates are for the first corner of the pixel
+        (the one nearest the [0, 0] corner of the tile in data space.
+        """
+        out = np.zeros(self.expected_data_shape + (3,), dtype=np.float32)
+
+        # Prepare some arrays to use inside the loop
+        pixel_ss_index, pixel_fs_index = np.meshgrid(
+            np.arange(0, self.frag_ss_pixels),
+            np.arange(0, self.frag_fs_pixels),
```
antarcticrainforest: You might want to add dtypes to the np.arange calls to avoid incompatible type errors

karabo_data/geometry2.py
```diff
@@ -335,6 +335,68 @@ def to_distortion_array(self):
...
+            indexing='ij'
+        )
+        if centre:
+            pixel_ss_index += 0.5
+            pixel_fs_index += 0.5
```
antarcticrainforest: I also don't understand this, pixel_ss/fs_indexes are the pixels/points in ss and fs direction on a 2D grid? Then why adding half a pixel in each direction? I mean correct me if I am wrong but shouldn't the pixels represent the smallest addressable elements on a lattice? And hence are the center by default ... well unless there is a fixed axis associated with them. I'll think about it

karabo_data/geometry2.py
```diff
@@ -335,6 +335,68 @@ def to_distortion_array(self):
...
+
+        The output is an array with shape like the data, with an extra dimension
+        of length 3 to hold (x, y, z) coordinates. Coordinates are in pixel
+        units from the detector centre.
+
+        If centre=True, the coordinates are calculated for the centre of each
+        pixel. If not, the coordinates are for the first corner of the pixel
+        (the one nearest the [0, 0] corner of the tile in data space.
+        """
+        out = np.zeros(self.expected_data_shape + (3,), dtype=np.float32)
```
antarcticrainforest: Wouldn't it be better to have a 2D array as input, e.g an array of definitions of multiple ROI's, and map that 2D array onto the 3D array for (nmodules x slow_scan x fast_scan). I know this is quite different to the approach below but maybe worthwhile considering?

karabo_data/geometry2.py
takluyver: Yep, good call.

karabo_data/geometry2.py
takluyver: If you imagine a Cartesian grid, the first pixel stretches from x=0 to x=1, and y=0 to y=1. Its indices are (0, 0), but the coordinates of its centre is (0.5,  0.5).

Calling these variables `index` is wrong, though - that was lazy copy-pasting.

karabo_data/geometry2.py
takluyver: I understand the desire to 'disassemble' an image, and I can implement that if required. But I think this approach is actually quite powerful - you can use the full precision and 3D information of the available geometry information to define a mask.

This kind of question is why I'm waiting for Henry's input before adding tests and docs for what I've implemented.

antarcticrainforest: Ok good, we'll keep that in mind at leave it as it is for now

antarcticrainforest:
LGTM, given that Henry is ok with it.

fangohr:
Do we have a notebook as an example that could go into the Karabo data documentation? I think this is interesting enough to be on read the docs. (Thinking in particular about the discussion with Andreas and the general need of having to know where pixels are for checking and integration/feeding of other tools etc).

takluyver:
I've just added a bit to the DSSC geometry notebook that was already there.

karabo_data/geometry2.py
```diff
@@ -500,6 +493,128 @@ def to_distortion_array(self):
...
+        same shape. module_no should hold integers, starting from 0,
+        so 0: Q1M1, 1: Q1M2, etc.
+
+        slow_scan and fast_scan describe positions within that module.
+        They may hold floats for sub-pixel positions. In both, 0.5 is the centre
+        of the first pixel.
+
+        Returns an array of similar shape with an extra dimension of length 3,
+        for (x, y, z) coordinates in metres.
+        """
```
tmichela: I do not understand this method and what should be the inputs :(

karabo_data/geometry2.py
```diff
@@ -500,6 +493,128 @@ def to_distortion_array(self):
...
+                )
+                pixels_y = (
+                        corner_y
+                        + pixel_ss_coord * ss_unit_y
+                        + pixel_fs_coord * fs_unit_y
+                )
+                pixels_z = (
+                        corner_z
+                        + pixel_ss_coord * ss_unit_z +
+                        + pixel_fs_coord * fs_unit_z
```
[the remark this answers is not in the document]
takluyver: So was I, but then I remembered that + and - can both be unary operators, i.e. `a + (+b)`. I'll remove it.

karabo_data/geometry2.py
```diff
@@ -500,6 +493,128 @@ def to_distortion_array(self):
...
+        if centre:
+            # A pixel is from n to n+1 in each axis, so centres are at n+0.5.
+            pixel_ss_coord += 0.5
+            pixel_fs_coord += 0.5
+
+        for m, mod in enumerate(self.modules, start=0):
+            for t, tile in enumerate(mod, start=0):
+                corner_x, corner_y, corner_z = tile.corner_pos * self.pixel_size
+                ss_unit_x, ss_unit_y, ss_unit_z = tile.ss_vec * self.pixel_size
+                fs_unit_x, fs_unit_y, fs_unit_z = tile.fs_vec * self.pixel_size
```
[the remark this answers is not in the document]
takluyver: I do consider a box around the pixel, so if you pass `centre=False`, it will give a point just outside the pixel (the corner of the bounding box). I'm OK with that. I'm planning to implement the distortion array method based on this, and that uses offsets from the corner of the bounding box.

However, I have made a mistake and got the centres slightly off. We should be adding 2/3 instead of 1/2 in the slow-scan direction - for the same reason as the 4/3 multiplication in the other PR you reviewed recently.

karabo_data/geometry2.py
takluyver: I'll try to clarify it tomorrow. But briefly, it's for things like the output of the robust peak finder: it gives peak locations in pixel coordinates within the input data - e.g. 31.2 pixels in the slow-scan direction, 10.78 pixels fast scan. This method can convert those into coordinates in physical space (x, y, z).

fangohr:
Two more things: 

Context: the plotting of the hexagons and the pixel positions is really great and useful in understanding and debugging things.

1. Can you create yet another plot of this type that shows (parts of) (at least) two quadrants together? In particular, I want to emphasise the point that the relative x and y positions of these quadrants will only depend on the quadrant position parameters, and thus can be arbitrary. 

I imagine this will also be useful when we evaluate various 'snapping' algorithms: we will then want to plot the 'assumed' pixel position on top of the real (hexagonal) pixel position. 

2. Can you extend this to also plot the 'assumed' positions for the `plot_fast` method? Ideally, we would draw the rectangular box of that pixel over the true hexagonal pixel. If we can see how this changes from one quadrant to the next, it would be perfect as a base for future development. 

If it is too hard to squeeze in now, then please object and convert into a new ticket. I think we'll need it in any case.

karabo_data/geometry2.py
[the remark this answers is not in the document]
takluyver: To illustrate what I meant about the centres being off, here's the plot before the correction:

pixel-centres-before2

And here it is after replacing the 1/2 shift with 2/3:

pixel-centres-after

takluyver:
- Labelling axes: done
- Reversed x: it's a consequence of the XFEL convention that we draw detector images looking along the beam, not into the beam.
- Plotting centres and hexagons together: I've done this by adding an `allow_negative_xy` parameter to `to_distortion_array`, rather than showing the transformation in the notebook - I think it would be more confusing than useful.
- Plotting multiple quadrants, rectangles over hexagons: I would recommend that we open new issues and tackle these separately. This PR is already pretty big, and I want to get a new release out soon. I'm also don't entirely follow what you want to show with parts of multiple quadrants.

takluyver:
The second new method is now called `data_coords_to_positions()`, and I've tried to document it better, including an example of using it in the AGIPD geometry notebook.

takluyver:
Got an offline LGTM from @tmichela

PR #147, opened 2019-04-15
Open run by proposal and run number

takluyver:
Add a new `open_run` function which can look up a run's data by proposal and run number:

```python
run = open_run(proposal=2012, run=238)
```

The proposal is looked up by scanning the instrument and cycle directories; this wastes a bit of time, but probably not much. The user can pass a proposal directory path to avoid this. And we could implement a caching mechanism later if it's an issue.

By default, it tries to combine data from raw and proc, with data in proc taking precedence where possible. This can be overridden with `proc=False` to only use raw data.

I'm not 100% sure about the way raw data is selected when combining raw+proc. Currently, it excludes any raw files for which the 3rd part of the filename (like `AGIPD05` or `DA02`) matches something in proc. But if e.g. one AGIPD module doesn't get calibrated for some reason, it will load raw data for that module and proc data for the others, which will cause problems if you later try to combine them. I don't think there's any way around that without hardcoding some specific rules about AGIPD/LPD/DSSC.

### DIFF
--- a/karabo_data/read_machinery.py
+++ b/karabo_data/read_machinery.py
@@ -3,14 +3,21 @@
 The public API is in karabo_data.reader; this is internal code.
 """
 from collections import defaultdict
+from glob import iglob
+import logging
 import numpy as np
 import os.path as osp
 import re
+import time
 
 from .exceptions import SourceNameError
 
+log = logging.getLogger(__name__)
+
 DETECTOR_NAMES = {'AGIPD', 'LPD'}
 DETECTOR_SOURCE_RE = re.compile(r'(.+)/DET/(\d+)CH')
+
+DATA_ROOT_DIR = '/gpfs/exfel/exp'
 
 
 class _SliceConstructor(type):
@@ -183,3 +190,18 @@ class FilenameInfo:
             )
         else:
             self.description = "Unknown data source ({})", datasrc
+
+
+def find_proposal(propno):
+    """Find the proposal directory for a given proposal on Maxwell"""
+    if '/' in propno:
+        # Already passed a proposal directory
+        return propno
+
+    t0 = time.monotonic()
+    for d in iglob(osp.join(DATA_ROOT_DIR, '*/*/{}'.format(propno))):
+        dt = time.monotonic() - t0
+        log.info("Found proposal dir %r in %.2g s", d, dt)
+        return d
+
+    raise Exception("Couldn't find proposal dir for {!r}".format(propno))
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -1063,7 +1063,7 @@ def RunDirectory(path):
 RunHandler = RunDirectory
 
 
-def open_run(proposal, run, proc=True):
+def open_run(proposal, run, data='raw'):
     """Access data from a specified run in the EuXFEL files on Maxwell.
 
     Parameters
@@ -1073,12 +1073,9 @@ def open_run(proposal, run, proc=True):
         '/gpfs/exfel/exp/SPB/201701/p002012'.
     run: str, int
         A run number such as 243, '243' or 'r0243'.
-    proc: bool
-        If True (default), open files from the 'proc' (processed) data folder,
-        and raw data files where there isn't a corresponding processed file.
-        Files which do not need post-processing are not copied to proc, so
-        this combination is necessary to include all data sources for the run.
-        If False, open only the raw data files.
+    data: str
+        'raw' or 'proc' (processed) to access data from one of those folders.
+        The default is 'raw'.
     """
     if isinstance(proposal, int):
         proposal = 'p{:06d}'.format(proposal)
@@ -1092,31 +1089,7 @@ def open_run(proposal, run, proc=True):
     elif not run.startswith('r'):
         run = 'r' + run.rjust(4, '0')
 
-    if not proc:
-        return RunDirectory(osp.join(prop_dir, 'raw', run))
-
-    raw_dir = osp.join(prop_dir, 'raw', run)
-    proc_dir = osp.join(prop_dir, 'proc', run)
-
-    # Not using glob here because it suppresses errors when you don't have
-    # permission to list the files.
-    raw_files = [e.path for e in os.scandir(raw_dir)
-                 if e.is_file() and e.name.endswith('.h5')]
-    proc_files = [e.path for e in os.scandir(proc_dir)
-                 if e.is_file() and e.name.endswith('.h5')]
-
-    # Names look like RAW-R0243-AGIPD10-S00002.h5 . If the 3rd part ('AGIPD10')
-    # exists in proc, skip loading it from raw.
-    got_in_proc = {osp.basename(p).split('-')[2] for p in proc_files}
-    raw_files = [p for p in raw_files
-                 if osp.basename(p).split('-')[2] not in got_in_proc]
-
-    files = proc_files + raw_files
-    if not files:
-        raise Exception("No .h5 files found in {!r} or {!r}"
-                        .format(raw_dir, proc_dir))
-
-    return DataCollection.from_paths(files)
+    return RunDirectory(osp.join(prop_dir, data, run))
 
 
 def stack_data(train, data, axis=-3, xcept=()):
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -19,10 +19,8 @@ import logging
 import numpy as np
 import os.path as osp
 import pandas as pd
-from pathlib import Path
 import re
 import sys
-import time
 from warnings import warn
 import xarray
 
@@ -35,6 +33,7 @@ from .read_machinery import (
     _tid_to_slice_ix,
     union_selections,
     contiguous_regions,
+    find_proposal,
 )
 
 __all__ = [
@@ -1063,21 +1062,6 @@ def RunDirectory(path):
 RunHandler = RunDirectory
 
 
-def find_proposal(propno):
-    """Find the proposal directory for a given proposal on Maxwell"""
-    if '/' in propno:
-        # Already passed a proposal directory
-        return propno
-
-    t0 = time.monotonic()
-    for d in Path('/gpfs/exfel/exp').glob('*/*/{}'.format(propno)):
-        dt = time.monotonic() - t0
-        log.info("Found proposal dir %r in %.2g s", d, dt)
-        return d
-
-    raise Exception("Couldn't find proposal dir for {!r}".format(propno))
-
-
 def open_run(proposal, run, proc=True):
     """Access data from a specified run in the EuXFEL files on Maxwell.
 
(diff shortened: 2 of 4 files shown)


karabo_data/reader.py
```diff
@@ -1058,6 +1062,52 @@ def RunDirectory(path):
...
+    Parameters
+    ----------
+    proposal: str, int
+        A proposal number, such as 2012, '2012', 'p002012', or a path such as
+        '/gpfs/exfel/exp/SPB/201701/p002012'.
+    run: str, int
+        A run number such as 243, '243' or 'r0243'.
+    proc: bool
+        If True (default), combine data from the raw and 'proc' (processed)
+        folders for the specified run. If False, access only the raw data.
```
tmichela: Can we precise here that the corrected file is loaded if it exists in both raw and proc directories

karabo_data/read_machinery.py
```diff
@@ -183,3 +190,18 @@ def __init__(self, path):
...
+def find_proposal(propno):
+    """Find the proposal directory for a given proposal on Maxwell"""
+    if '/' in propno:
+        # Already passed a proposal directory
+        return propno
+
+    t0 = time.monotonic()
+    for d in iglob(osp.join(DATA_ROOT_DIR, '*/*/{}'.format(propno))):
+        dt = time.monotonic() - t0
+        log.info("Found proposal dir %r in %.2g s", d, dt)
```
tmichela: this is only for debug, right?

tmichela: what's the overhead?

karabo_data/reader.py
```diff
@@ -1058,6 +1062,52 @@ def RunDirectory(path):
...
+    # Names look like RAW-R0243-AGIPD10-S00002.h5 . If the 3rd part ('AGIPD10')
+    # exists in proc, skip loading it from raw.
+    got_in_proc = {osp.basename(p).split('-')[2] for p in proc_files}
+    raw_files = [p for p in raw_files
+                 if osp.basename(p).split('-')[2] not in got_in_proc]
+
+    files = proc_files + raw_files
+    if not files:
+        raise Exception("No .h5 files found for proposal {!r} run {!r}"
+                        .format(proposal, run))
```
tmichela: I wonder if it would be more helpful here to show the resolved path

karabo_data/reader.py
```diff
@@ -1058,6 +1062,52 @@ def RunDirectory(path):
...
+    proc: bool
+        If True (default), combine data from the raw and 'proc' (processed)
+        folders for the specified run. If False, access only the raw data.
+    """
+    if isinstance(proposal, int):
+        proposal = 'p{:06d}'.format(proposal)
+    elif ('/' not in proposal) and not proposal.startswith('p'):
+        proposal = 'p' + proposal.rjust(6, '0')
+
+    prop_dir = find_proposal(proposal)
```
tmichela: I wonder if we should add an additional test in case the user has not read access to the proposal, because right now the error message is `no .h5 files found in ...` which is not helpful, maybe something like:

```python
if not os.access(prop_dir, os.R_OK):
    raise Exception(f'cannot access {prop_dir}: permission denied')
```

karabo_data/reader.py
takluyver: Yep, good idea. Update coming.

karabo_data/read_machinery.py
takluyver: The overhead of the logging should be minimal - this is what the logging framework is for, after all. I just timed an unhandled (not printing anything) `log.info()` call at 400 ns.

I'm starting to use logging also in PR #150. We might make the logging visible in some command-line wrappers, and we might ask people to enable it when they run into problems. In this case, I want to be ready to keep an eye on how long it takes to find proposals in case we need to figure out some caching later.

karabo_data/reader.py
takluyver: Agreed

karabo_data/reader.py
takluyver: Rather than adding an explicit test, I've switched from `glob()` to `os.scandir()` (which is like listdir but a bit fancier). `glob()` suppresses errors encountered while it's scanning, whereas using the lower level `os` functions means a standard PermissionError should be raised.

karabo_data/read_machinery.py
tmichela: Ah sorry, I meant: what is the overhead of scanning into the file system?

takluyver: Good question. I don't know; I'm expecting it to be fairly low at the moment, compared to the cost of opening dozens of HDF5 files once the proposal is found. But it will presumably grow as the number of cycle directories grows.

tmichela:
One last comment I have: I am a bit afraid that people use this, getting `proc` data, do their processing, everything works... They re-process it a few weeks later, but the `proc` data is removed in the meantime and this will silently give them raw data leading to either unexpected results or breaking their code.

I wonder if, for example, it would be better to feedback (error, warning, something else?) when `proc` flag is true and no `proc` data is found? Maybe I'm too cautious... any opinion?

fangohr:
I agree with Thomas M. The proc and raw data is fundamentally different and automatic (transparent) substitution of one with the the other is likely to lead to wrong results or confusion and delays. 

This will become more pronounced as processed data is meant to be created on demand and then deleted again after a short time - like one or two weeks. 

I think it is better if the user needs to say what they want (at least for detectors that have a  alit still routine) and if it is not available, an error should be raised. 

We can later  tailor that error to suggest that missing processed data can be created by going to the metadata catalog.

takluyver:
I think the current code will fail in that case when it tries to examine the proc run directory, but that's definitely accidental, not planned. And I'm assuming the directory is removed, not just emptied.

Should we offer a (possibly default) option to read only `proc` files? That would then error if they were deleted because it would find no usable files. I made combining raw+proc the default to facilitate things like correlating detector data with e.g. motor positions, but in practice the analyses we've come across so far seem to use either raw or proc, not both together, so maybe I'm letting neat ideas beat practicality.

fangohr:
I would also expect that the time will come when fast 2d detector data needs to be processed making use of slow data (which would be from the raw files). 

Is the path (within the h5 file) the same for raw and processed data? If so, maybe we should introduce an additional qualifier that makes this unique? Like „:proc“ or such?

tmichela:
Yes, the run directory is deleted and not emptied.
The current implementation is probably ok then of it fails when the directory does not exists on proc (with meaningful error message) ...

> Should we offer a (possibly default) option to read only proc files? That would then error if they were deleted because it would find no usable files.

That's maybe the safest solution at the moment (with an explicit proc+raw option).

Alternatively, what about adding to the calibration routine the new write_virtual feature including all files untouched by calibration. So that only raw or proc dir would be independently opened?

takluyver:
> Is the path (within the h5 file) the same for raw and processed data? If so, maybe we should introduce an additional qualifier that makes this unique? Like „:proc“ or such?

It is the same. It might make sense for the calibration to produce output with a different name, at least for AGIPD where the calibrated data is qualitatively different from the raw data. Possibly it makes less sense when the calibration is just things like background subtraction.

However, renaming it would also break existing code that expects it at the current names. And I'm reluctant to try some fake-renaming in karabo_data, because that's error prone and makes it harder to understand what is really in the files.

> > Should we offer a (possibly default) option to read only proc files? That would then error if they were deleted because it would find no usable files.
> 
> That's maybe the safest solution at the moment (with an explicit proc+raw option).

OK, I'll look into that. Any preference between these API ideas:

```python
open_run(..., data_from="raw+proc")
open_run(..., data_from=('raw', 'proc')
open_run(..., raw=True, proc=True)
```

> Alternatively, what about adding to the calibration routine the new write_virtual feature including all files untouched by calibration. So that only raw or proc dir would be independently opened?

That would be convenient from our perspective at the moment, but I'm not sure it's conceptually right to expose all the raw data in proc. 

It could also be a practical problem in the future. If there were two pipelines processing different parts of the raw data and writing different files into proc, which is responsible for building the virtual file with the remainder of data sources?

tmichela:
> It could also be a practical problem in the future. If there were two pipelines processing different parts of the raw data and writing different files into proc, which is responsible for building the virtual file with the remainder of data sources?

good point!

> OK, I'll look into that. Any preference between these API ideas:
```
open_run(..., data_from="raw+proc")
open_run(..., data_from=('raw', 'proc')
open_run(..., raw=True, proc=True)
```

Any of these is ok for me. Originally I thought of something like the first one, but I find the 3nd a bit neater. And looking at the 2nd I imagine it would also be nice to have possibility to open several runs, like:
```python
run = open_run(proposal=2012, run=[238, 250, 265], ...)
run = open_run(proposal=2012, run=[238:250], ...)
```
is this going too far? :D

takluyver:
Thinking about it more, I suspect it's tricky to make the `raw=True, proc=True` option intuitive, because it's easy to assume that `raw=True` means you only get raw data. Longer names like `include_raw` could make it clearer... but also longer. So maybe that points to one of the other options.

We could also do multiple runs easily enough, but I'd say let's leave that for a separate PR.

fangohr:
I prefer long names for clarity over short names ("include_raw"?). [I am not sure about the implications of the settings, but trust your judgment.]

Regarding the observation that the calibrated data is under the same path as the raw data: I think that's a design that we need to discuss and question more widely. Of course one can argue that the directory and file path contains that information, but through karabo_data we abstract away the different directory paths etc, so I think within karabo_data we must have some way of using a unique path (or path and annotation). [I can't see the value of substituting processed data with raw data if the processed data is missing; I can only see confusion and non-reproducibility arise from that.]

takluyver:
There's no intention to substitute processed data with raw data - it's just a potential side-effect of how we decide which raw data to include.

Maybe until we can work out a better answer for this, we should drop the 'combining' part, and just provide this as a convenient way to open a raw OR proc run directory?

takluyver:
I went to remove the combining option, but I don't think there's all that much use in the function without it. So maybe we should leave this PR on hold until we have an answer we're happy with to that.

takluyver:
I also have a concern about what happens if calibration has failed for some modules of a big detector. The way this selects the files to open means it will load a mixture of raw and processed data from the detector, and then something confusing will happen if you try to stack them together. We could maybe work around this by hardcoding more specific knowledge about filenames and detectors into it, but that doesn't seem like a very reliable answer.

If we were to remove the part about combining raw+proc, is being able to use a proposal & run number rather than a run directory path still worth pursuing by itself?

tmichela:
I think that it is still useful. At least I would use it.
I often end up typing the whole `/gpfs/exfel/.../.../.../.../p001234/proc/r0000` and I'd be happy to type instead (and this is pure laziness): `open_run(1234, 0000)`

takluyver:
OK, I've done that.

Do you think raw or proc should be the default? I've currently got raw as the default, because most of the sources aren't in proc. You'd specify `data='proc'` to get proc instead.

tmichela:
LGTM!

PR #148, opened 2019-04-16
Jungfrau geom

dallanto:
Just a formal pull request for reviewing the bit of added code, @takluyver, @tmichela

### DIFF
--- a/README_jungfrau_geom.md
+++ b/README_jungfrau_geom.md
@@ -4,3 +4,5 @@ This starts with a single module with its 2 x 4 tiles as geometry fragments, in 
 
 Each Jungfrau ASIC has a pixel margin of double physical size. This fact is ignored at the moment, like for the other detectors, meaning that the border pixel rows/columns are treated like normal pixels concerning size and photon counts. The only implication is that tile origins in the composed image must account for the resulting 2 px ASIC gap.
 
+With respect to AGIPD, we have a swapping of (ss, fs) from (x,y) to (y,x).
+
--- a/README_jungfrau_geom.md
+++ b/README_jungfrau_geom.md
@@ -1,0 +1,6 @@
+# Adding a geometry description for the Jungfrau detector
+
+This starts with a single module with its 2 x 4 tiles as geometry fragments, in analogy to AGIPD and LPD.
+
+Each Jungfrau ASIC has a pixel margin of double physical size. This fact is ignored at the moment, like for the other detectors, meaning that the border pixel rows/columns are treated like normal pixels concerning size and photon counts. The only implication is that tile origins in the composed image must account for the resulting 2 px ASIC gap.
+
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -1,4 +1,4 @@
-"""AGIPD & LPD geometry handling."""
+"""AGIPD & LPD geometry handling; Jungfrau added 2019/04/11"""
 from cfelpyutils.crystfel_utils import load_crystfel_geometry
 from copy import copy
 import h5py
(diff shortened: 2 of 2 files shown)


takluyver:
Thanks, this looks good. I'm happy to see that the geometry code makes enough sense that someone else can extend it.

Have you tested all the methods available from DetectorGeometryBase? `inspect()`, `position_modules_fast()`, `plot_data_fast()` and `to_distortion_array()`?

We'll also need to integrate it into the documentation (see `docs/geometry.rst`).

dallanto:
Hi Thomas,
systematic testing of the methods you mention has yet to be done. I will do so soon, however I guess I will need extra information from multi-module data (or geometry files etc.) in order to test e.g. `position_modules_fast()`. The principal problem at the moment is that we have to wait for concrete examples from either EuXFEL or PSI/SwissFEL (I am in contact with the JF developers).

dallanto:
Anyway, thanks Thomas for looking into my code extensions.

takluyver:
Sorry Fabio, this rather got sidelined in the drive to integrate DSSC geometry. I'd like to aim for a release now that that has landed, but then we should come back to this.

takluyver:
@dallanto - are you still interesting in picking this up again? Someone was asking about assembling Jungfrau images recently on Redmine, so it might be relevant.

dallanto:
> @dallanto - are you still interesting in picking this up again? Someone was asking about assembling Jungfrau images recently on Redmine, so it might be relevant.

Hi Thomas,
yes I am still interested. I just need to coordinate/prioritize this wrt my other tasks, as there are now requirements from both SPB and MID. I've actually seen the Redmine ticket, but concerning geometry descriptions for multi-module cases, I've not received any news (initially I've been in contact with PSI people). What do you think would be a reasonable starting point?

takluyver:
Ah, I'd forgotten that this was describing a single module. I'd say it's reasonable to start with something similar, with a parameter for the gap between modules, and `expected_data_shape = (2, 512, 1024)`. Then someone can adjust that gap parameter to make a preview look right.

takluyver:
PS be aware when you update it that - among other changes - the internals of the geometry machinery now work in metres, not pixel units.

dallanto:
> Ah, I'd forgotten that this was describing a single module. I'd say it's reasonable to start with something similar, with a parameter for the gap between modules, and `expected_data_shape = (2, 512, 1024)`. Then someone can adjust that gap parameter to make a preview look right.

Ok so you suggest with "something similar" to create a separate class named e.g. "Jungfrau_1M" having the shape with extra dimension, and the gap parameter?

takluyver:
Yup, that sounds good to me.

dallanto:
Btw. do you have a clue about the branch conflict message according to which we would require command-line usage for resolving?

takluyver:
I guess the changes on master (renaming and splitting up the file) make the conflicts too much for Github to resolve them in the web UI.

You can try to resolve it locally, but as you're adding a completely new class, it might be easier to start a new branch from master and copy/paste the new code from this.

dallanto:
> I guess the changes on master (renaming and splitting up the file) make the conflicts too much for Github to resolve them in the web UI.
> 
> You can try to resolve it locally, but as you're adding a completely new class, it might be easier to start a new branch from master and copy/paste the new code from this.

Yes, I cloned the 0.6.0-tagged repository and created a new "jungfrau1M_geom" branch from master, then I copied the JF_500K class code into the geometry2/__init__.py and also started a separate class JF_1M as first commit (all steps local so far). I will continue on this tomorrow.

takluyver:
:+1: thanks

dallanto:
> PS be aware when you update it that - among other changes - the internals of the geometry machinery now work in metres, not pixel units.

Hmm, I have the inspect method as part of the JF_500KGeometry class, which uses the one of the base class via super().inspect(...). There, in the base method, I find the default parameter axis_units='px'. With that, the axes are labeled 'Pixels' and the scale is to 10^7. So that the panel tags (A0 to A7) all fall onto the origin.
Now when I call the method from my JF class by super().inspect(axis_units='m'), the scaling all becomes more reasonable, with correctly placed panel tags. However: the axis labels are 'metres' (makes sense), but the scale is to 1024 x 512 which are the pixel dimensions. Somehow the two units do not fit yet ...

takluyver:
Yup, when you create the instance, you'll need to convert `corner_pos`, `ss_vec` and `fs_vec` from pixels to metres, i.e. multiply by the pixel size.

PR #149, opened 2019-04-16
Clarify that write_virtual is not like hdf5-virtualise

takluyver:
Add a note clarifying what `write_virtual` doesn't do.


tmichela:
LGTM :)

PR #152, opened 2019-04-29
added notebook file for XGM two runs case

dallanto:
@takluyver
This is the pre-discussed notebook on XGM. I keep it in its own, newly created branch, but would ask you to review it once more. Thus, the pull request. Thank you.


takluyver:
This looks good. A few notes that come to mind looking through it:

1. The title refers to 'separate runs'. That sounds a bit like different points in time. I'd say something like 'separate SASEs' or 'different beamlines'
2. How do you know that these two runs will overlap? Is it possible to find this from the metadata catalogue? Even if the answer is just 'ask XPD', it would be good to mention in the introduction who knows.
3. I think the code selecting the first pulses could use a bit more explanation. `sa1_flux[:,0]` is very convenient, but a bit cryptic if people aren't already familiar with Python/Numpy. And it might be surprising if they're coming from a language with 1-indexing. We should explain that the first dimension is pulse trains, the second is individual pulses, and that `[:, 0]` selects everything in the train dimension and the first value in the pulse dimension.
4. Let's avoid using the shorthand `add_subplot(111)` - either use [fig.subplots](https://matplotlib.org/api/_as_gen/matplotlib.figure.Figure.html#matplotlib.figure.Figure.subplots) or spell out `add_subplot(1, 1, 1)` instead. Packing three logically separate numbers into one is convenient but doesn't make anything clearer.

Finally, if you wanted to extend this, we could use an [xarray dataset](http://xarray.pydata.org/en/stable/data-structures.html#dataset), which is like a multidimensional equivalent of a pandas dataframe. So we could align and join the two XGM datasets in full, and then select the first pulses only at the last step for plotting.

takluyver:
Oh, also the tests are failing because nbval attempts to run all the notebooks, and it can't load the data on Travis. There's an exclusion list in `pytest.ini`. We should probably switch to explicitly including the notebooks which can run without the data.

dallanto:
Thanks for looking into this. I agree to all (1)-(4). In fact, if you remember the session, I did not apply the numpy [:, 0] option, but did something like `data_first_pulse = np.array([item[0] for item in data])` actually even with for-loop and append command. I could either switch back to something like that or, preferably keep it with numpy index notation but explain in more detail, like you do.
I will also explore your suggestion at the end of the comment.

takluyver:
I like the numpy syntax - it's really convenient for things like this once you're used to it, and it's much faster than using a Python loop. We want the example to show people the best way of doing things. So let's leave it in and just expand the explanation a bit. :-)

dallanto:
@takluyver 
I just pushed a new version addressing your points.
1. Maybe the title is still not optimal, but I changed to emphasize the simultaneous nature of this experiment. I feel that the title will get too long if also the spatial seperation is put into it. The additional text in the title cell make it clear.
2. I now mentioned the XPD group there as well. To me, this now reads as it should, i.e. that the a priori information about synchronicity comes from them (otherwise one would acknowledge this as soon as the trainIds are compared.)
3. Numpy syntax for "first-pulses across trains" selection: as discussed
4. Changed subplot allocation as suggested, but no extra comment on this.

dallanto:
About CI: do I understand correctly that - because this notebook *cannot* run without the data - we should put it into the exclusion list in order to avoid the check failure?

takluyver:
That's right about CI.

It would also be good to add the notebook to the list of examples in `index.rst`, so it's included as part of the docs.

takluyver:
You've still got some `add_subplot(111)` in there; otherwise I think this is looking promising.

dallanto:
I added the notebook file name to the ignore-list (without quotes, with or w/o both seems o.k.)
About the index.rst: I think this is the plan but I want to get Hans' confirmation today at the sitdown if we really want to include it to the docs.

takluyver:
LGTM

PR #155, opened 2019-05-03
DSSC geometry

takluyver:
This is largely based on guesswork for which way round things should go; we'll need to use it with some data to verify it, but I wanted to make a start.

dssc_inspect

As best I could tell, the geometry file we got was again opposite to the standard XFEL axis directions, like the original LPD geometry. What I've committed here is the inverse of that - i.e. this file matches the standard axis directions, if I've understood things correctly.


karabo_data/geometry2.py
```diff
@@ -1055,8 +1055,114 @@ def invert_xfel_lpd_geom(path_in, path_out):
...
+class DSSC_Geometry(DetectorGeometryBase):
+    """Detector layout for DSSC
+
+    The coordinates used in this class are 3D (x, y, z), and represent multiples
+    of the pixel size.
+
+    You won't normally instantiate this class directly:
+    use one of the constructor class methods to create or load a geometry.
+    """
+    # Hexagonal pixels, 236 μm step in slow-scan axis, 204 μm in fast-scan
```
antarcticrainforest: So prefer approximating the hexagons to squares? We could also assume that the pixels are 236x272 (the pointy bit inclusive)

takluyver: I'm still unsure about exactly how to represent it in the geometry machinery. But if we are approximate them as rectangular pixels for drawing, the important dimension is really the step distance from one pixel to the next, rather than the size across an individual pixel. That makes it easy to calculate an approximately accurate position for each pixel.

takluyver:
Just to record, this is my current best guess of which way round the modules are with respect to the coordinate system, and which point the geometry file is measuring to:

dssc_guesswork

It's tricky to check any of this without a complete frame of data we can view.

takluyver:
The code can now assemble data by 'snapping' to a regular grid, with no interpolation, but handling the fact that pixels do not have a 1:1 aspect ratio.

I've now guessed the opposite data orientation to before. I have no new information,  but the inspect plot makes slightly more sense than my previous one. We will still need to figure out what order the data is recorded in before this is usable.

Current layout guess:

dssc_guesswork

dssc_inspect2


Example of assembling empty image:

dssc_shaped

takluyver:
For the record: this is not quite as clear as I'd like, because the geometry code is based on pixel units, which aren't clear when the pixels don't have a 1:1 aspect ratio. I want to change this (#160), but I don't plan to do it as part of this PR.

antarcticrainforest:
Since this branch has now already been used 'in production' for quite some time I'd say we can merge this: LGTM

takluyver:
Thanks. I'll try to write a few tests for it. By which I mean copy the tests for LPD and modify them.

PR #156, opened 2019-05-09
Improve docs on selecting data

takluyver:


karabo_data/reader.py
```diff
@@ -653,21 +653,44 @@ def select(self, seln_or_source_glob, key_glob='*'):
...
+
         3. With a dict of source names mapped to sets of key names
            (or empty sets to get all keys)::
 
+            # Select image.data from one detector source, and all data from one XGM
             sel = run.select({'SPB_DET_AGIPD1M-1/DET/ALLCH:xtdf': {'image.data'},
                               'SA1_XTD2_XGM/XGM/DOOCS': set()})
 
+           Unlike the others, this option *doesn't* allow glob patterns.
+           It's a more advanced option for code that knows precisely what
```
tmichela: What does "more advanced" means here? is it faster to use it that way?
Also the `ALLCH:xtdf` in the example confuses me a bit... is that a thing?

takluyver: It's more advanced in the sense that it's precise - you'll never unexpectedly get sources that you didn't mean to select - but less convenient, because you have to spell out the source names you want in full.

I think the ALLCH bit must have been copied across from a karabo-bridge example at some point - it's not a thing in the files, as far as I know. I'll change that.

takluyver: (This option is likely to be marginally faster, because the other options are expanded into this structure, but the difference should be small enough not to matter)

tmichela:
ok, thanks, LGTM.

takluyver:
Thanks. I'll leave it a little while for Hans to have a look, as he was asking for clarifications. Otherwise, I'll merge it some time next week.

PR #159, opened 2019-05-13
Document checking and working around issues with the data

takluyver:
As requested by email. The bottom of the 'Reading data files' page was the best place I could think of to add this.

I also noticed that one documented method was missing a docstring, so I added that.


tmichela:
Very nice, thanks, LGTM

PR #161, opened 2019-05-21
added notebook version approved by Jan ... publications pending

dallanto:
@takluyver 
Hi Thomas. Just in order to not forget it: this is the version from last week that has found approval by Jan. I open the pull request so that you may look into the new code (the parts concerning error calculation), if you haven't done so from the e-mail attachments.
I guess we won't merge to master anyway before the citations are complete. And only then, I guess, would an updated index.rst make sense (note that the notebook has a new name wrt. last version: "xgm_two_runs" => "xpd_examples2").
Best, Fabio


dallanto:
Yes, and just now I also added the name to the exlusion list for CI   : )

pytest.ini
```diff
@@ -1,2 +1,3 @@
 [pytest]
-addopts = --ignore docs/apply_geometry.ipynb --ignore docs/xpd_examples.ipynb --ignore "docs/agipd_geometry.ipynb" --ignore docs/xgm_two_runs.ipynb
+addopts = --ignore docs/apply_geometry.ipynb --ignore docs/xpd_examples.ipynb --ignore "docs/agipd_geometry.ipynb" --ignore docs/xgm _two_runs.ipynb
```
takluyver: You've accidentally introduced a space in `xgm_two_runs`

takluyver:
Is this building on the content in the `xgm_two_runs` notebook? Would it make sense to extend/replace that rather than adding a separate example?

takluyver:
The content mostly looks good at a glance. But can we find a nice consistent way for matplotlib to show train IDs as X-axis labels? I don't think they should be appearing with a *1e7* multiplier in the corner. Either each label should have a full train ID (maybe at an angle), or show the last few digits with a *+38227000* addition in the corner.

When we merge it will be up to XPD. Do you know how far along the papers they want us to cite are? Is there an 'in press' or Arxiv version we could already point to?

pytest.ini
dallanto: oops, indeed thanks for correcting that

dallanto:
> Is this building on the content in the `xgm_two_runs` notebook? Would it make sense to extend/replace that rather than adding a separate example?

I shall not become a separate example, nor extent. Replacement is intended. Which means, of course, that we could/should remove the "xgm_two_runs" altogether

pytest.ini
dallanto: Well, I did now

dallanto:
Finally CI worked, after I removed the old version both as file from the repo and as entry from the exclusion list. Thus, pls. ignore my intermediate answer. There has not only been the space within the name, but also an unintended line-break, as it seemed.

dallanto:
About the line-plot x-axis with train identifiers: Thanks for the suggestion on label text, I will look into that.

dallanto:
> When we merge it will be up to XPD. Do you know how far along the papers they want us to cite are? > Is there an 'in press' or Arxiv version we could already point to?

Theo from XPD wrote:
"we have 2 new XGM papers which are "almost" published.
 Please use:
 A. A. Sorokin et al., J. Synchrotron Rad. 26, ????? (2019) [XT5016]
 Th. Maltezopoulos et al., J. Synchrotron Rad. 26, ????? (2019) [XT5015]
 The ????? will soon be available, just remove ????? for the moment."

One might replace the ????? with "in press" or remove as suggested. I've found nothing in arXiv. I guess I will ask first when they expect the two articles to be out.

takluyver:
Thanks, I think the train ID tick labels look good like that. So this is ready to go from our side whenever XPD are happy about the references.

dallanto:
> Thanks, I think the train ID tick labels look good like that. So this is ready to go from our side whenever XPD are happy about the references.

Indeed, thanks for giving feedback!

takluyver:
I think these papers are public now, but they still don't have a page number that I can see:

- Sorokin et al.: https://onlinelibrary.wiley.com/iucr/doi/10.1107/S1600577519005174
- Maltezopoulos et al.: https://onlinelibrary.wiley.com/doi/10.1107/S1600577519003795

takluyver:
+1 - let's also make the DOIs links to the relevant papers. You can prepend `https://dx.doi.org/10.foo` to any DOI to get a link that will redirect to it.

takluyver:
LGTM once the tests pass

PR #162, opened 2019-05-28
'On demand' virtual datasets

takluyver:
This adds a `data.get_virtual_dataset(source, key)` method, similar to `get_array()`, which creates and returns a virtual dataset for a single data field. This doesn't load the data into memory, so it's practical for accessing large amounts of data.

For instance, you could create a virtual dataset, see the shape of the data as if it was one giant array, and then tell multiple worker processes to each process a different chunk of that data.

(The virtual datasets for now are just stitching together sequence files for a single data source, not joining multiple detector modules. But we could add a similar interface for that later.)

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -22,6 +22,7 @@ import os.path as osp
 import pandas as pd
 import re
 import sys
+import tempfile
 from warnings import warn
 import xarray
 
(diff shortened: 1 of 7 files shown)


karabo_data/reader.py
```diff
@@ -994,6 +995,47 @@ def write_virtual(self, filename):
...
+        data processing, you can copy the temporary file to another location.
+        It should be a small file, as it only points to the real data. However,
+        it will only be usable where you have access to the original data files.
+        """
+        self._check_field(source, key)
+
+        from .writer import VirtualFileWriter
+        fd, filename = tempfile.mkstemp(suffix='-karabo-data-vds.h5')
+        os.close(fd)
+        vfw = VirtualFileWriter(filename, self)
```
tmichela: Is there something preventing us to use the `core` driver + no backing store? That could be used to avoid creating a file at all (file would exists only in memory)

takluyver: I'm considering it a feature that there's a real file. My thinking is that you can pass the path and dataset name to worker processes, which would open the file to access the data.

I'm considering having an optional `filename=` parameter to specify where the file should be created, and using the tempfile if it's not given.

antarcticrainforest: +1 for the in memory option using the core driver. I had some cases where users wanted to use virtual datasets to do easier processing of detector data across trains. So I guess creating a memory view of a virtual dataset could be of value. Think about it.

takluyver: There's not much user-visible difference between a purely in-memory object and creating a small temp file to hold the virtual dataset. Since we want a real file for some use cases anyway, I'm inclined to stick with the default file driver, and avoid any obscure bugs that might come from using the core driver.

antarcticrainforest:
LGTM!

takluyver:
I'll add the (optional) `filename=` parameter and then ask you to take another quick look, if that's OK.

Also, don't you have today off? Stop working! :stuck_out_tongue:

takluyver:
I've added the `filename=` parameter, so I'm happy with this now.

takluyver:
I've written a notebook example showing how you can use this to split data up for multiple workers to process.

takluyver:
Got an LGTM from @fangohr on Slack.

PR #166, opened 2019-06-05
Quick fix: Check if data.mask exists

antarcticrainforest:
We ran into the problem of write_cxi failing because the LPD data has no mask entry. This very minor PR should check if a mask is existing.


takluyver:
Just to check, this was calibrated data? It had `gain` but not `mask`?

antarcticrainforest:
Correct!

takluyver:
OK, thanks. :-)

PR #168, opened 2019-06-05
Move crystFEL related methods to base class.

antarcticrainforest:
This PR aims at loading/writing CFEL geometry file format for all detectors.

Background is that CFEL people planning an experiment at FXE with the LPD. Hence they want to use crystfel with LPD data.

### DIFF
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -277,40 +277,40 @@ class DetectorGeometryBase:
             path to the group that contains the mask array in the hdf5 file
 
         dims : collection, default : ('frame', 'modno', 'ss', 'fs')
-            Dimensions of the data. Extra dimensios, except for details are
-            added by thier index. The dimensions should contain frame, modno,
+            Dimensions of the data. Extra dimensions, except for the defaults,
+            should be added by therr index e.g ('frame', 'modno', 0, 'ss', 'fs)
+            for raw data. Note: the dimensions must contain frame, modno,
             ss, fs.
 
-        adu_per_ev : float, default : 0.0075
+        adu_per_ev : float
             ADU (analog digital units) per electron volt for the considered
             detector.
 
-        clen : float, default : 0.119
+        clen : float
             Distance between sample and detector in meters
 
-        photon_energy : int, default : 9800
+        photon_energy
             Beam wave length in eV
         """
         from . import __version__
 
         if adu_per_ev is None:
-            warnings.warn('adu_per_ev must be set; Setting to default value')
             adu_per_ev_str = '; adu_per_eV = SET ME'
             # TODO: adu_per_ev should be fixed for each detector, we should 
             #       find out the values and set them.
         else:
             adu_per_ev_str = 'adu_per_eV = {}'.format(adu_per_ev)
+
         if clen is None:
             clen_str = '; clen = SET ME'
-            warnings.warn('clen must be set; Setting to default value')
         else:
             clen_str = 'clen = {}'.format(clen)
+
         if photon_energy is None:
             photon_energy_str = '; photon_energy = SET ME'
-            warnings.warn('photon_energy must be set, setting to default value')
         else:
             photon_energy_str = 'photon_energy = {}'.format(photon_energy)
-        
+
         # Get the frame dimension
         tile_dims = {}
 
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -246,8 +246,24 @@ class DetectorGeometryBase:
         rigid_string += 'rigid_group_collection_asics = {}\n\n'.format(modules)
         return rigid_string
 
-    def write_crystfel_geom(self, filename):
-        """Write this geometry to a CrystFEL format (.geom) geometry file."""
+    def write_crystfel_geom(self, filename, adu_per_ev=0.0075, clen=0.119,
+                            photon_energy=10235):
+        """Write this geometry to a CrystFEL format (.geom) geometry file.
+        Parameters
+        ----------
+        filename : str
+        filename of the geometry file
+
+        Keywords
+        --------
+        adu_per_ev : float
+        ADU (analog digital units) per electron volt for the considered 
+        detector (default 0.0075).
+        clen : float
+        Distance between sample and detector in meters (default 0.119)
+        photon_energy : int
+        Beam wave length in eV (default 10235)
+        """
         from . import __version__
 
         panel_chunks = []
@@ -257,7 +273,10 @@ class DetectorGeometryBase:
         pix_size = self.pixel_size * 1e6 # Convert pixels size to microns
         with open(filename, 'w') as f:
             f.write(CRYSTFEL_HEADER_TEMPLATE.format(version=__version__,
-                                                    pix_size=pix_size))
+                                                    pix_size=pix_size,
+                                                    adu_per_ev=adu_per_ev,
+                                                    clen=clen,
+                                                    photon_energy=photon_energy))
             rigid_groups = self._get_rigid_groups()
             f.write(rigid_groups)
             for chunk in panel_chunks:
@@ -804,9 +823,9 @@ CRYSTFEL_HEADER_TEMPLATE = """\
 
 dim0 = %
 res = {pix_size} ;
-photon_energy = 10235 ;
-clen = 0.119  ; Camera length, aka detector distance
-adu_per_eV = 0.0075  ; no idea
+photon_energy = {photon_energy} ;
+clen = {clen}  ; Camera length, aka detector distance
+adu_per_eV = {adu_per_ev}; no idea
 
 """
 
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -217,7 +217,7 @@ class DetectorGeometryBase:
 
         return ax
 
-    def _tile_dims(self, modno, tileno):
+    def _tile_dims(self, tileno):
         """Implement in subclass: which part of module array each tile is.
         """
         raise NotImplementedError
@@ -278,7 +278,7 @@ class DetectorGeometryBase:
 
         dims : collection, default : ('frame', 'modno', 'ss', 'fs')
             Dimensions of the data. Extra dimensions, except for the defaults,
-            should be added by therr index e.g ('frame', 'modno', 0, 'ss', 'fs)
+            should be added by their index e.g ('frame', 'modno', 0, 'ss', 'fs)
             for raw data. Note: the dimensions must contain frame, modno,
             ss, fs.
 
@@ -325,7 +325,7 @@ class DetectorGeometryBase:
         panel_chunks = []
         for p, module in enumerate(self.modules):
             for a, fragment in enumerate(module):
-                ss_dims, fs_dims = self._tile_dims(p, a)
+                ss_dims, fs_dims = self._tile_dims(a)
                 panel_chunks.append(fragment.to_crystfel_geom(p,
                                                               a,
                                                               ss_dims,
@@ -774,11 +774,7 @@ class AGIPD_1MGeometry(DetectorGeometryBase):
         return ss_slice, fs_slice
 
     @classmethod
-    def _tile_dims(cls, modno, tileno):
-        # Which part of the array is this tile?
-
-        #min_ss=(a * self.ss_pixels),
-        #max_ss=(((a + 1) * self.ss_pixels) - 1),
+    def _tile_dims(cls, tileno):
         tile_offset = tileno * cls.frag_ss_pixels
         ss_dims = tile_offset, tile_offset + cls.frag_ss_pixels - 1
         fs_dims = 0, cls.frag_fs_pixels - 1 # Every tile covers the full pixel range
@@ -1164,7 +1160,7 @@ class LPD_1MGeometry(DetectorGeometryBase):
         return super().to_distortion_array()  # Overridden only for docstring
 
     @classmethod
-    def _tile_dims(cls, moduleno, tileno):
+    def _tile_dims(cls, tileno):
         if tileno < 8: # First half of module (0 <= t <=7)
             fs_dims = 0, 127
             tiles_up = 7 - tileno
@@ -1354,7 +1350,7 @@ class DSSC_Geometry(DetectorGeometryBase):
         return ss_slice, fs_slice
 
     @classmethod
-    def _tile_dims(cls, modno, tileno):
+    def _tile_dims(cls, tileno):
         tile_offset = tileno * cls.frag_fs_pixels
         fs_dims = tile_offset, tile_offset + cls.frag_fs_pixels - 1
         ss_dims = 0, cls.frag_ss_pixels - 1# Every tile covers the full pixel range
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -278,9 +278,9 @@ class DetectorGeometryBase:
 
         dims : collection, default : ('frame', 'modno', 'ss', 'fs')
             Dimensions of the data. Extra dimensions, except for the defaults,
-            should be added by their index e.g ('frame', 'modno', 0, 'ss', 'fs)
-            for raw data. Note: the dimensions must contain frame, modno,
-            ss, fs.
+            should be added by their index, e.g.
+            ('frame', 'modno', 0, 'ss', 'fs') for raw data.
+            Note: the dimensions must contain frame, modno, ss, fs.
 
         adu_per_ev : float
             ADU (analog digital units) per electron volt for the considered
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -70,6 +70,7 @@ class GeometryFragment:
             p=p,
             min_ss=(a * self.ss_pixels),
             max_ss=(((a + 1) * self.ss_pixels) - 1),
+            max_fs=self.fs_pixels,
             ss_vec=_crystfel_format_vec(self.ss_vec),
             fs_vec=_crystfel_format_vec(self.fs_vec),
             corner_x=c[0],
@@ -816,7 +817,7 @@ CRYSTFEL_PANEL_TEMPLATE = """
 {name}/dim3 = fs
 {name}/min_fs = 0
 {name}/min_ss = {min_ss}
-{name}/max_fs = 127
+{name}/max_fs = {max_fs}
 {name}/max_ss = {max_ss}
 {name}/fs = {fs_vec}
 {name}/ss = {ss_vec}
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -145,6 +145,8 @@ class DetectorGeometryBase:
     pixel_size = 0.0
     frag_ss_pixels = 0
     frag_fs_pixels = 0
+    n_modules = 0
+    n_asics_per_module = 0
     expected_data_shape = (0, 0, 0)
     _pixel_shape = np.array([1., 1.])  # Overridden for DSSC
     _draw_first_px_on_tile = 1  # Tile num of 1st pixel - overridden for LPD
@@ -204,6 +206,40 @@ class DetectorGeometryBase:
             ax.invert_xaxis()
 
         return ax
+
+    @classmethod
+    def from_crystfel_geom(cls, filename):
+        """Read a CrystFEL format (.geom) geometry file.
+
+        Returns a new geometry object.
+        """
+        geom_dict = load_crystfel_geometry(filename)
+        modules = []
+        for p in range(cls.n_modules):
+            tiles = []
+            modules.append(tiles)
+            for a in range(cls.n_asics_per_module):
+                d = geom_dict['panels']['p{}a{}'.format(p, a)]
+                tiles.append(GeometryFragment.from_panel_dict(d))
+        return cls(modules, filename=filename)
+
+    def write_crystfel_geom(self, filename):
+        """Write this geometry to a CrystFEL format (.geom) geometry file."""
+        from . import __version__
+
+        panel_chunks = []
+        for p, module in enumerate(self.modules):
+            for a, fragment in enumerate(module):
+                panel_chunks.append(fragment.to_crystfel_geom(p, a))
+        pix_size = self.pixel_size * 1e6 # Convert pixels size to microns
+        with open(filename, 'w') as f:
+            f.write(CRYSTFEL_HEADER_TEMPLATE.format(version=__version__,
+                                                    pix_size=pix_size))
+            for chunk in panel_chunks:
+                f.write(chunk)
+
+        if self.filename == 'No file':
+            self.filename = filename
 
     def _snapped(self):
         """Snap geometry to a 2D pixel grid
@@ -369,6 +405,8 @@ class AGIPD_1MGeometry(DetectorGeometryBase):
     frag_ss_pixels = 64
     frag_fs_pixels = 128
     expected_data_shape = (16, 512, 128)
+    n_modules = 16
+    n_asics_per_module = 8
 
     @classmethod
     def from_quad_positions(cls, quad_pos, asic_gap=2, panel_gap=29,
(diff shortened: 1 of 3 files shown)


karabo_data/geometry2.py
```diff
@@ -772,7 +778,11 @@ def plot_data(self, modules_data, axis_units='px', frontview=True):
 ; See: http://www.desy.de/~twhite/crystfel/manual-crystfel_geometry.html
 
 dim0 = %
-res = 5000 ; 200 um pixels
+res = {pix_size} ;
+photon_energy = 10235 ;
+clen = 0.119  ; Camera length, aka detector distance
+adu_per_eV = 0.0075  ; no idea
```
takluyver: I don't particularly like the idea of hardcoding these parameters into the geometry file.

I've been meaning for a while to make them optional arguments for `write_crystfel_geom()` so you can supply them when writing the file. If you want to tackle that as part of this PR, that would be fine. Otherwise, feel free to leave them out of this PR and we can get to it later.

karabo_data/geometry2.py
```diff
@@ -205,6 +207,40 @@ def inspect(self, frontview=True):
...
+        return cls(modules, filename=filename)
+
+    def write_crystfel_geom(self, filename):
+        """Write this geometry to a CrystFEL format (.geom) geometry file."""
+        from . import __version__
+
+        panel_chunks = []
+        for p, module in enumerate(self.modules):
+            for a, fragment in enumerate(module):
+                panel_chunks.append(fragment.to_crystfel_geom(p, a))
```
takluyver: The creation of the chunk for an individual panel is a bit AGIPD-specific at the moment - in particular around the min/max slow-scan and fast-scan coordinates defining where the panel is.

karabo_data/geometry2.py
antarcticrainforest: Agreed! I added some keywords to the methods. That are passed into the header!

karabo_data/geometry2.py
antarcticrainforest: Yes! I added some additional keywords to the fragment header. I've tested it with all three available detectors. It seems to work quite fine.

I am inclined to say that the CFEL format should become our default format and that we should push users using this format (or at least encourage them). I think this PR would be a great step into this direction.

antarcticrainforest:
I am not sure why the tests are failing but I'll try to fix this next week after I'll return from my leave in Rügen!

karabo_data/geometry2.py
```diff
@@ -70,6 +70,7 @@ def to_crystfel_geom(self, p, a):
             p=p,
             min_ss=(a * self.ss_pixels),
             max_ss=(((a + 1) * self.ss_pixels) - 1),
+            max_fs=self.fs_pixels,
```
takluyver: Unfortunately this isn't quite sufficient. This was written for AGIPD, where each tile spans the full fast-scan dimension (i.e. `min_fs == 0 and max_fs == (fs_pixels - 1)`). But in DSSC for instance, each tile spans the full slow-scan dimension, but only half of the fast-scan dimension.

As part of #142, I added a `_tile_slice` method on classes which would define what section of the data (in fast-scan & slow-scan dimensions) a given tile corresponds to. We'd probably want to use that or something similar to determine the min/max fs/ss for CrystFEL.

karabo_data/geometry2.py
```diff
@@ -205,6 +208,83 @@ def inspect(self, frontview=True):
...
+        for p in range(cls.n_modules):
+            tiles = []
+            modules.append(tiles)
+            for a in range(cls.n_asics_per_module):
+                d = geom_dict['panels']['p{}a{}'.format(p, a)]
+                tiles.append(GeometryFragment.from_panel_dict(d))
+        return cls(modules, filename=filename)
+
+    def _get_rigid_groups(self, nquads=4):
+        """Create ridget stings for rigid groups definiton."""
```
takluyver: ridget?

antarcticrainforest:
I am not sure why the CI is still failing. The error message is rather convoluted.

karabo_data/geometry2.py
```diff
@@ -205,6 +208,83 @@ def inspect(self, frontview=True):
...
+            rigid_string += 'rigid_group_p{} = {}\n'.format(nn, ','.join(rigid_group_p))
+
+        rigid_string += '\n'
+
+        rigid_string += 'rigid_group_collection_quadrants = {}\n'.format(quads)
+        rigid_string += 'rigid_group_collection_asics = {}\n\n'.format(modules)
+        return rigid_string
+
+    def write_crystfel_geom(self, filename, adu_per_ev=0.0075, clen=0.119,
+                            photon_energy=10235):
```
takluyver: I'd rather not have default values here. Presumably they will rarely be right, especially across different detectors and instruments, but it's very easy for someone who doesn't know the real values to assume that the defaults are probably a good starting point. They might either waste time trying to do something that can't work, or even produce invalid results because they've got bad parameters.

Of course it's ultimately the scientist's responsibility to notice things like this and find the correct values. But our software tools should make it easy to do the right thing and hard to do the wrong thing.

I don't think the karabo_data API should require these parameters - just have `None` for the defaults and write them into the geometry file only if they're specified.

karabo_data/geometry2.py
```diff
@@ -205,6 +208,83 @@ def inspect(self, frontview=True):
...
+        """Write this geometry to a CrystFEL format (.geom) geometry file.
+        Parameters
+        ----------
+        filename : str
+        filename of the geometry file
+
+        Keywords
+        --------
```
takluyver: The convention I'm used to is to list all these under 'Parameters', and either describe them as `float, optional`, or just rely on the description to note that they're optional.

karabo_data/geometry2.py
takluyver: I'd also make them keyword-only arguments, which works like this:

https://github.com/European-XFEL/karabo_data/blob/9728c5b9be6fbbec6ccbc4ce09d99421e6ba56ba/karabo_data/geometry2.py#L816-L817

karabo_data/geometry2.py
```diff
@@ -145,6 +146,8 @@ class DetectorGeometryBase:
     pixel_size = 0.0
     frag_ss_pixels = 0
     frag_fs_pixels = 0
+    n_modules = 0
+    n_asics_per_module = 0
```
takluyver: I'm generally moving towards using the XFEL terminology ('tiles' instead of 'asics') in this code. Some of the CrystFEL geometry code will continue to match the CrystFEL terms, but for this new attribute, I think it's best to use 'tiles'.

('asics' may also be somewhat inaccurate; I think I saw that the tiles we position for DSSC actually contain multiple ASICs)

karabo_data/geometry2.py
```diff
@@ -205,6 +208,83 @@ def inspect(self, frontview=True):
...
+            rigid_string += 'rigid_group_q{} = {}\n'.format(nn, ','.join(rigid_group_q))
+        rigid_string += '\n'
+        for nn, rigid_group_p in enumerate(np.array_split(rigid_group, self.n_modules)):
+            rigid_string += 'rigid_group_p{} = {}\n'.format(nn, ','.join(rigid_group_p))
+
+        rigid_string += '\n'
+
+        rigid_string += 'rigid_group_collection_quadrants = {}\n'.format(quads)
+        rigid_string += 'rigid_group_collection_asics = {}\n\n'.format(modules)
+        return rigid_string
```
takluyver: It would be good to have a bit of testing for this. It should be run by existing tests, but they don't check whether the results make sense.

antarcticrainforest: Not sure if enough though, but if've added some additional testing into existing test methods.

karabo_data/geometry2.py
antarcticrainforest: so n_modules and n_tiles_per_module?

karabo_data/geometry2.py
antarcticrainforest: Agreed!

karabo_data/geometry2.py
antarcticrainforest: I guess rigid

karabo_data/geometry2.py
antarcticrainforest: Yes I've added a method to get the dimensions of a tile

karabo_data/geometry2.py
antarcticrainforest: okidoke

karabo_data/geometry2.py
takluyver: Yup :-)

karabo_data/geometry2.py
takluyver: Thanks, that looks good.

karabo_data/geometry2.py
```diff
@@ -205,6 +216,114 @@ def inspect(self, frontview=True):
...
+                            adu_per_ev, clen, photon_energy):
+        """Write this geometry to a CrystFEL format (.geom) geometry file.
+        Parameters
+        ----------
+        data_path  : (str)
+            path to the group that contains the data array in the hdf5 file
+            (default : /entry_1/instrument_1/detector_1/data)
+        mask_path : (str)
+            path to the group that contains the mask array in the hdf5 file
+            (default : /entry_1/instrument_1/detector_1/mask)
```
takluyver: There is no default!

karabo_data/geometry2.py
```diff
@@ -205,6 +216,114 @@ def inspect(self, frontview=True):
...
+
+        extra_dim (dim_number, index) : collection
+            extra data dimension that has to inserted to read the data. The
+            first number of the tuple refers to the axis number; the second to
+            the index of that axis which holds the data of interest.
```
takluyver: I'm having some trouble understanding this. I guess it's to do with rearranging which axis means what, but the pairs of numbers feel kind of opaque. I'd suggest we leave the data shape hardcoded in this PR (i.e. dim1 = module no, dim 2 & 3 = fast-scan & slow-scan), and deal with making it more flexible as a separate change.

karabo_data/geometry2.py
```diff
@@ -205,6 +216,114 @@ def inspect(self, frontview=True):
...
+
+        rigid_string += '\n'
+
+        rigid_string += 'rigid_group_collection_quadrants = {}\n'.format(quads)
+        rigid_string += 'rigid_group_collection_asics = {}\n\n'.format(modules)
+        return rigid_string
+
+    def write_crystfel_geom(self, data_path='/entry_1/instrument_1/detector_1/data',
+                            mask_path=None, extra_dim=None, *, filename,
+                            adu_per_ev, clen, photon_energy):
```
takluyver: To avoid breaking the API, the filename needs to be the first parameter, and any parameters we're adding need to be optional  - at the moment this would require `adu_per_ev`, `clen` and `photon_energy`, because they don't have any defaults.

karabo_data/geometry2.py
```diff
@@ -1050,6 +1129,15 @@ def to_distortion_array(self):
         """
         return super().to_distortion_array()  # Overridden only for docstring
 
+    @classmethod
+    def _tile_dims(cls, tileno):
+        # Which part of the array is this tile?
+        tile_offset = tileno * cls.frag_ss_pixels
+        ss_dims = tile_offset, tile_offset + cls.frag_ss_pixels - 1
+        fs_dims = 0, cls.frag_fs_pixels - 1 # Every tile covers the full pixel range
+        return ss_dims, fs_dims
```
takluyver: This doesn't look right for LPD - there's a 2 x 8 tile layout, so it doesn't cover the full range in either dimension. Here's the `_tile_slice` method in #142:

https://github.com/European-XFEL/karabo_data/blob/f14ebb4f6e6ad35bc6c3740bbe6ba9d7cd11ad3b/karabo_data/geometry2.py#L1142-L1153

The tile layout within the module is in the top-right of this diagram:

lpd_layout

karabo_data/geometry2.py
```diff
@@ -62,14 +62,23 @@ def centre(self):
...
-    def to_crystfel_geom(self, p, a):
+    def to_crystfel_geom(self, p, a, ss_dims, fs_dims, extra_dim):
         name = 'p{}a{}'.format(p, a)
         c = self.corner_pos
+        dims = [p, 'ss', 'fs']
+        if extra_dim:
+            dims.insert(extra_dim[0]-1, extra_dim[1])
+        dims_str = ''
+        for num, value in enumerate(dims):
+            dims_str += '{}/dim{} = {}\n'.format(name, num+1, value)
```
takluyver: I think this will also go slightly wrong,  because it will start at `dim0` instead of `dim1`. As I mentioned in another comment, I think we should leave this particular bit of flexibility for a separate PR.

karabo_data/geometry2.py
antarcticrainforest: Right

karabo_data/geometry2.py
antarcticrainforest: well what I meant was the following:
Suppose you have raw data which has the shape of (n_frames, n_modules, 2, fast_scan, slow_scan). If you wanted to get the data you'd access it via the second index of the 3rd dimension. I guess that would be expressed in terms of crystfel geometry code:

```a0p0/dim1 = 0
   a0p0/dim2 = 1
   a0p0/dim3 = fs
   a0p0/dim4 = ss
```
I did this in order to make the hdfsee read the dssc data. In my test this was working fine. To be honest I don't think it's too complicated so I guess we do not have to do this in a seperate PR. Maybe it's just a matter of how it is explained in the doc string

karabo_data/geometry2.py
antarcticrainforest: yes correct, I'll do that!

karabo_data/geometry2.py
antarcticrainforest: I should start with dim1 as I add num+1.

karabo_data/geometry2.py
antarcticrainforest: Ok I'll fix this

karabo_data/geometry2.py
takluyver: Yes, my mistake!

karabo_data/geometry2.py
takluyver: OK, I'll have another look tomorrow morning to see if it makes sense to me then, and/or if I can see how to clarify the code at all.

takluyver: What if instead of a parameter to insert an extra dimension, the parameter is a tuple describing all the dimensions, with a default like `('modno', 'ss', 'fs')`? Then to handle an extra dimension for raw data, you would put `dims=('modno', 0, 'fs', 'ss')`.

From a usage perspective, it would be even nicer if it could start from dim0, like `('%', 'modno', 'ss', 'fs')` or `('frame', 'modno', 'ss', 'fs')`. But that might make the implementation trickier, because at the moment `dim0` is described in the header for all panels at once.

takluyver: I guess we could start from `dim0` but require that it's the frame number for now.

`%` is CrystFEL's own placeholder for this in the geometry file, so using that makes it less magic, but obviously something like `frame` is more explicit.

karabo_data/geometry2.py
```diff
@@ -205,6 +216,123 @@ def inspect(self, frontview=True):
...
+
+        if adu_per_ev is None:
+            adu_per_ev = 0.0075
+            warnings.warn('adu_per_ev must be set; Setting to default value')
+        if clen is None:
+            clen = 0.119
+            warnings.warn('clen must be set; Setting to default value')
+        if photon_energy is None:
+            photon_energy = 9800
+            warnings.warn('photon_energy must be set, setting to default value')
```
takluyver: I'd rather just write these commented out (`; photon_energy =`), or with a nonsense value that has to be replaced before you can use it with CrystFEL. People generally ignore warnings, especially if the warning isn't clear what needs to be done.

karabo_data/geometry2.py
```diff
@@ -771,44 +881,20 @@ def plot_data(self, modules_data, axis_units='px', frontview=True):
...
 
 """
 
 
 CRYSTFEL_PANEL_TEMPLATE = """
-{name}/dim1 = {p}
-{name}/dim2 = ss
-{name}/dim3 = fs
-{name}/min_fs = 0
+{dims}{name}/min_fs = {min_fs}
```
takluyver: I'd ideally like to arrange the logic so `{dims}` can go on its own line, for clarity. It should be easy enough with a `"\n".join(...)`.

antarcticrainforest: Ok, will do!

karabo_data/geometry2.py
antarcticrainforest: Yes you're right! The comment idea (;) is a good one. This can also be useful once we offer beam_energy values as part of the virtual datasets and do not need to use the values defined here!

karabo_data/geometry2.py
antarcticrainforest: Thats a good idea, I'll look into that tmr!

karabo_data/geometry2.py
antarcticrainforest: Done!

karabo_data/geometry2.py
antarcticrainforest: I am not 100% sure about the dimensions for LPD though. To be honest I do not 100% understand the code to get the dimensions for the distortion. Maybe you might want to have a look and tell me if that's ok. I could not check with hdfsee because of the issue @ebadkamil posted. But since this is solved now maybe @ebadkamil could try creating a VDS again ;) ?

ebadkamil: There is now a new virtual dataset that has all the 16 modules. Same location in scratch folder.

karabo_data/geometry2.py
```diff
@@ -1050,6 +1138,22 @@ def to_distortion_array(self):
...
+        # Which part of the array is this tile?
+        module_offset = moduleno * 256
+        if tileno < 8: # First half of module (0 <= t <=7)
+            fs_dims = 0, 127
+            tiles_up = 7 - tileno
+        else:
+            fs_dims = 128, 255
+            tiles_up = tileno - 8
+
+        tile_offset = module_offset + (tiles_up * 32)
```
takluyver: I don't think you need to use `module_offset` for this. The virtual datasets we generate have a separate array dimension for the modules (panels in CrystFEL terminology). So the slow-scan/fast-scan coordinates we record are only within each module (0 to 255 in both dimensions for LPD).

antarcticrainforest: Ok!

karabo_data/geometry2.py
antarcticrainforest: Yes, that seems to work! Thx @ebadkamil

karabo_data/geometry2.py
antarcticrainforest: I also guess adu_per_ev is detector specific, what about giving a default number per detector and falling back to this default value if not specified?

karabo_data/geometry2.py
antarcticrainforest: Ok I think I am pretty happy with this solution for now.

karabo_data/geometry2.py
antarcticrainforest: All three detectors seem to work.

karabo_data/geometry2.py
takluyver: If we can get good numbers and be confident that they're ~fixed for each detector, great. But unless you've got those numbers to hand already, I'd say leave it with no default for now, and we can work that out separately later.

karabo_data/geometry2.py
```diff
@@ -205,6 +217,139 @@ def inspect(self, frontview=True):
...
+            Distance between sample and detector in meters
+
+        photon_energy : int, default : 9800
+            Beam wave length in eV
+        """
+        from . import __version__
+
+        if adu_per_ev is None:
+            warnings.warn('adu_per_ev must be set; Setting to default value')
+            adu_per_ev_str = '; adu_per_eV = 0.0075'
```
takluyver: If we're having no default, I'd not even write a number in the commented line. Just put `; adu_per_eV = XXX` or `; adu_per_eV = SET ME` or something. Otherwise, people will just uncomment it with the value that's already there. Then we can get rid of the warnings as well.

karabo_data/geometry2.py
```diff
@@ -1050,6 +1161,22 @@ def to_distortion_array(self):
         """
         return super().to_distortion_array()  # Overridden only for docstring
 
+    @classmethod
+    def _tile_dims(cls, moduleno, tileno):
+        # Which part of the array is this tile?
+        module_offset = moduleno * 256
```
takluyver: This line is no longer needed.

karabo_data/geometry2.py
antarcticrainforest: Yes that makes sense. But lets try not to forget about that. Maybe I'll add a TODO comment.

karabo_data/geometry2.py
antarcticrainforest: Ok!

karabo_data/geometry2.py
antarcticrainforest: Can you cross check?

karabo_data/geometry2.py
```diff
@@ -205,6 +217,141 @@ def inspect(self, frontview=True):
...
+        filename : str
+        filename of the geometry file
+
+        data_path  : str, default : /entry_1/instrument_1/detector_1/data
+            path to the group that contains the data array in the hdf5 file
+        mask_path : str
+            path to the group that contains the mask array in the hdf5 file
+
+        dims : collection, default : ('frame', 'modno', 'ss', 'fs')
+            Dimensions of the data. Extra dimensios, except for details are
```
takluyver: dimensios -> dimensions

takluyver: Also, I'm not clear what "except for details" refers to.

karabo_data/geometry2.py
```diff
@@ -205,6 +217,141 @@ def inspect(self, frontview=True):
...
+            ss, fs.
+
+        adu_per_ev : float, default : 0.0075
+            ADU (analog digital units) per electron volt for the considered
+            detector.
+
+        clen : float, default : 0.119
+            Distance between sample and detector in meters
+
+        photon_energy : int, default : 9800
```
takluyver: These three no longer have defaults

karabo_data/geometry2.py
```diff
@@ -205,6 +217,141 @@ def inspect(self, frontview=True):
...
+            Beam wave length in eV
+        """
+        from . import __version__
+
+        if adu_per_ev is None:
+            warnings.warn('adu_per_ev must be set; Setting to default value')
```
takluyver: I'd drop these warnings now. People don't have to set these, and warnings get annoying, so I'd rather reserve them for when something could really be a problem.

karabo_data/tests/test_agipd_geometry.py
```diff
@@ -22,23 +24,25 @@ def test_write_read_crystfel_file(tmpdir):
...
-    # We need to add some experiment details before cfelpyutils will read the
-    # file
-    with open(path, 'r') as f:
-        contents = f.read()
-    with open(path, 'w') as f:
-        f.write('clen = 0.119\n')
-        f.write('adu_per_eV = 0.0075\n')
-        f.write(contents)
+    geom.write_crystfel_geom(filename=path, photon_energy=9000,
+                             adu_per_ev=0.0075, clen=0.2)
```
takluyver: It would be good to have a test with each of the 3 detector classes, even if it's pretty basic. And at least one of them should be missing the photon_energy, adu_per_ev, clen parameters, so it exercises the code paths for when they have no value.

karabo_data/geometry2.py
```diff
@@ -205,6 +217,141 @@ def inspect(self, frontview=True):
 
         return ax
 
+    def _tile_dims(self, modno, tileno):
```
takluyver: I think this doesn't actually need the module number - at least for now, every module has the same layout of tiles with respect to how the data is stored.

karabo_data/geometry2.py
antarcticrainforest: That was defaults. I'll add a more comprehensive description.

karabo_data/geometry2.py
antarcticrainforest: Yes, yes indeed!

karabo_data/geometry2.py
antarcticrainforest: Ok

karabo_data/geometry2.py
```diff
@@ -277,40 +277,40 @@ def write_crystfel_geom(self, filename, *,
             path to the group that contains the mask array in the hdf5 file
 
         dims : collection, default : ('frame', 'modno', 'ss', 'fs')
-            Dimensions of the data. Extra dimensios, except for details are
-            added by thier index. The dimensions should contain frame, modno,
+            Dimensions of the data. Extra dimensions, except for the defaults,
+            should be added by therr index e.g ('frame', 'modno', 0, 'ss', 'fs)
```
takluyver: therr -> their

Also, missing a final `'` after fs.

karabo_data/tests/test_agipd_geometry.py
antarcticrainforest: Do we have a geometry test for dssc?

takluyver: Ah, of course. We do in master, but this branch split before it was added. I'll merge this and add a test separately.

takluyver: In fact, I don't think it even really makes sense for DSSC, because for now CrystFEL assumes square pixels. Even if we ignore the hexagonal pixels, you still have to treat them as rectangles to get them in roughly the right position.

karabo_data/geometry2.py
```diff
@@ -205,6 +217,141 @@ def inspect(self, frontview=True):
...
+        panel_chunks = []
+        for p, module in enumerate(self.modules):
+            for a, fragment in enumerate(module):
+                ss_dims, fs_dims = self._tile_dims(a)
+                panel_chunks.append(fragment.to_crystfel_geom(p,
+                                                              a,
+                                                              ss_dims,
+                                                              fs_dims,
+                                                              tile_dims))
+        pix_size = self.pixel_size * 1e6 # Convert pixels size to microns
```
takluyver: While doing a final read-through, I spotted that this isn't right. CrystFEL expresses resolution in pixels per metre. I'm adding a fix for that.

Issue #172, opened 2019-06-12
Virtual dataset from Detector components skips missing modules


ebadkamil:
Creating virtual Data set using `write_virtual_cxi` leads to ndarray with shapes
(pulses, modules, py, px) which is perfect. However when Martin produce geometry file in crystfel format from xfel format it has information about all 16 modules. This virtual dataset then may lead to problem with crystfel because the missing modules will not be available in one big NDarray. 

```
In [55]: from karabo_data.components import LPD1M                                                                                   

In [56]: from karabo_data import RunDirectory, by_index, by_id                                                                      

In [57]: r =  RunDirectory("/gpfs/exfel/exp/XMPL/201750/p700000/proc/r0007")                                                        

In [58]: s = r.select_trains(by_id[1487289920:1487289921])                                                                          

In [59]: lpd = LPD1M(s)                                                                                                             

In [60]: lpd.write_virtual_cxi('lpd.cxi')                                                                                           

In [61]: from h5py import File                                                                                                      

In [62]: f = File('lpd.cxi')                                                                                                        

In [63]: f['entry_1/data_1/data'].shape                                                                                             
Out[63]: (30, 13, 256, 256)
```

takluyver:
Thanks. I think it should have 16 in the modules dimension even if some of them are missing. I'll try to figure out where it's going wrong.

ebadkamil:
Thanks Thomas

takluyver:
I've just remembered that there is a `module_identifier` dataset in the CXI files, which should give the module number for each slice in the relevant dimension. So it's still possible for tools to correctly interpret the data in a CXI file.

Since the virtual dataset shouldn't need any extra space for the empty parts, I'll make it always have space for 16 modules so it's easier to interpret.

PR #173, opened 2019-06-12
Always write virtual CXI file with space for 16 modules

takluyver:
Closes gh-172

cc @ebadkamil

### DIFF
--- a/karabo_data/components.py
+++ b/karabo_data/components.py
@@ -82,7 +82,9 @@ class MPxDetectorBase:
     """
 
     _source_re = re.compile(r'(?P<detname>.+)/DET/(\d+)CH')
-    module_shape = (0, 0)  # Override in subclass
+    n_modules = 16
+    # Override in subclass
+    module_shape = (0, 0)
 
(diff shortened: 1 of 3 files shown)


karabo_data/components.py
```diff
@@ -82,7 +82,9 @@ class MPxDetectorBase:
     """
 
     _source_re = re.compile(r'(?P<detname>.+)/DET/(\d+)CH')
-    module_shape = (0, 0)  # Override in subclass
+    n_modules = 16
```
antarcticrainforest: Shouldn't that be part of the individual detector classes? I mean there might be the case when a detector does not consist of 16 modules?

takluyver: I was thinking that it could be overridden by detectors with other numbers of modules, but as the 3 we have so far are all composed of 16 modules, it's an easy shortcut to have 16 as the default.

What do you reckon, should we set it on each of the subclasses to be more explicit?

antarcticrainforest: Yes that's a fair point. Let's keep it then. Other than that LGTM

PR #174, opened 2019-06-17
Document DSSC geometry and data access

takluyver:
Including some diagrams for the geometry.


antarcticrainforest:
Stupid question, is there any way to see this in action?

takluyver:
That's not a stupid question! :-)

The docs are only built automatically on master, and for the latest tagged version. In principle, we can enable any other branch on RTD, but it doesn't seem to have noticed that this branch exists, so I can't do that at the moment.

To try it locally, check out the branch, cd into the docs folder, and run `make html`. This will create a folder `_build/html` containing the built docs. You'll need to have Sphinx & nbsphinx installed (both Python packages), plus all the regular dependencies of karabo_data (it imports it to get docstrings).

antarcticrainforest:
Thanks for the info. I was assuming that it might be a bit more involved. I'll try that

docs/geometry.rst
```diff
@@ -71,6 +71,41 @@ which this geometry code can position independently.
 
    .. automethod:: from_h5_file_and_quad_positions
 
+   .. automethod:: to_distortion_array
+
+   .. automethod:: plot_data_fast
+
+   .. automethod:: position_modules_fast
+
+   .. automethod:: inspect
```
antarcticrainforest: Shouldn't this also mention the crystfel methods (read/write)?

docs/geometry.rst
```diff
@@ -71,6 +71,41 @@ which this geometry code can position independently.
...
+.. figure:: _static/dssc_hexes.png
+
+   Detail of hexagonal pixels in the corner of one DSSC module.
+
+.. autoclass:: DSSC_Geometry
+
+   .. automethod:: from_h5_file_and_quad_positions
+
+   .. automethod:: to_distortion_array
+
```
antarcticrainforest: Also here the crystfel methods?

docs/geometry.rst
```diff
@@ -71,6 +71,41 @@ which this geometry code can position independently.
...
+DSSC-1M consists of 16 modules of 128×512 pixels each.
+Each module is further subdivided into 2 sensor tiles,
+which this geometry code can position independently.
+
+.. figure:: _static/dssc_layout.png
+
+   The approximate layout of DSSC-1M, in a front view (looking along the beam).
+
+The pixels in each DSSC module are tesselating hexagons.
```
antarcticrainforest: tessellating: After living for more than 5 years in an English speaking country I am still learning vocabulary. I am sure I will have forgotten this word by tomorrow.  I think you could use the word _mosaic_, which I believe is more universal in other languages and makes it more unlikely that people like me stumble on it.

docs/geometry.rst
takluyver: Yes, good point. I'll add them.

docs/geometry.rst
takluyver: I'm inclined not to mention them for DSSC for now, because as far as I know, CrystFEL can't support even rectangular pixels (it has a single `res` parameter to represent pixel density), so we can't represent it in CrystFEL geometry even ignoring the hexagonal pixels.

docs/geometry.rst
takluyver: Maybe we can just leave out the word altogether? I don't think mosaic necessarily means tessellating - lots of mosaics have gaps between the tiles to make the desired shapes.

antarcticrainforest: Nah, that's fine. If you want to keep it just keep it

antarcticrainforest:
LGTM!

takluyver:
Thanks!

PR #175, opened 2019-06-18
Fix docstring for shape of DSSC distortion array

takluyver:
Thanks @dallanto for pointing out the mistake here.


takluyver:
This is a sufficiently straightforward fix that I'm just going to merge it.

Issue #164, opened 2019-06-03
DSSC: Plot pixel data at pixel position


fangohr:
It turns out that for debugging of geometry file, geometry file interpretation and calibration of quadrants, it is important to see exactly what our best understanding of the detector geometry (i.e. the positions of the pixels in real space) is. Even if this takes a long time to plot.

I think we need a function that creates a plot, which shows the value of a pixel at its position in real space as much as possible.

For example, we could place for every pixel a circle (dot) at the centre of the pixel location (and colour it according to value and colormap). One step up would be to draw the hexagonal shape of the pixels (this would required to draw 6 lines to form the hexagon, or a hexagonal polygon if that functionality is provided somewhere). 

For these plots I would not want to do any snapping (and thus we can't use `imshow` or such - we need to draw each pixel individually at the right coordinates.

It's likely to be slow and may need some manual fine-tuning to make it useful, and that's okay.

RobertRosca:
Hey, [here's the notebook I had](https://gist.github.com/RobertRosca/5aa6409511b614d858fcff46f8088862) which just plots the pixel positions based on distortion, either by fully drawing out the hexagon and using matplotlib's `fill` method (which draws polygons based on their corners), or by drawing a scatter plot of the centres of the hexagons

The fill method can take in a [colour option per polygon](https://matplotlib.org/3.1.0/api/_as_gen/matplotlib.pyplot.fill.html), so we could compute the RGB colour of each pixel based on their values then use that as the polygon colour

Similarily scatter can take in a [sequence of colours](https://matplotlib.org/3.1.0/api/_as_gen/matplotlib.pyplot.scatter.html) which would also work and be much faster than individually drawing each hex

PR #176, opened 2019-06-18
Fixes & example for DSSC distortion array

takluyver:
I played around with @RobertRosca's notebook for #164, and spotted that we were getting the hexagons rather wrong. I've added a bit based on this to the example in the docs,  and fixed the problems.

Here are the hexagons with the distortion array before the changes:

distortion-arr-before

And here they are after the changes:

distortion-arr-after


karabo_data/geometry2.py
```diff
@@ -1384,14 +1384,15 @@ def to_distortion_array(self):
...
+        # TODO: double check this shifting is correct
+        pixel_fs_index[1::2, :] -= 0.5
 
         # Corners described clockwise from the top, assuming the reference point
         # for a pixel is outside it, aligned with the top point & left edge.
-        corner_ss_offsets = np.array([0.5, 1, 1, 0.5, 0, 0])
-        corner_fs_offsets = np.array([0, 0.25, 0.75, 1, 0.75, 0.25])
+        # The 4/3 extends the hexagons into the next row to correctly tessellate.
+        corner_ss_offsets = np.array([0, 0.25, 0.75, 1, 0.75, 0.25]) * 4 / 3
+        corner_fs_offsets = np.array([0.5, 1, 1, 0.5, 0, 0])
```
tmichela: non expert in pyFAI here: Since the hexagons are stretched (204/236 I think), should this impact these corner positions? Or 1 unit in one axis does not necessarily equals 1 unit in the other axis?

Also is the fast scan the longest axis for DSSC?

takluyver: The 204/236 stretch is taken care of elsewhere - that's effectively the aspect ratio to get all pixels in (nearly) the right position.

But the 204 μm vertical distance is the step from the centre of one row of pixels to the centre of the next. If we just use that (i.e. remove the 4/3), then the top points of hexagons in row 0 are in line with the bottom points of row 1. We want them to interlock like the teeth of a gear. So the 4/3 is the factor between 204 and 272:

Pixel_dims

I did also check that 4/3 is what I'd expect for ideal regular hexagons.

The long axis (512 pixels) is indeed fast-scan for DSSC.

tmichela: aha, I understand. Thanks for explaining.

karabo_data/geometry2.py
```diff
@@ -1384,14 +1384,15 @@ def to_distortion_array(self):
             indexing='ij'
         )
         # Every second line of pixels across the slow-scan direction is shifted
-        # half a pixel in the fast-scan direction so the hexagons tessalate.
-        # TODO: check we're shifting the right way once we know the pixel order
-        pixel_fs_index[1::2, :] += 0.5
+        # half a pixel against the fast-scan direction so the hexagons tessalate.
+        # TODO: double check this shifting is correct
+        pixel_fs_index[1::2, :] -= 0.5
```
tmichela: This is just me thinking out loud trying to understand, please correct me if I'm wrong:

for right-hand side modules, the first pixel is at the bottom left.
It reads left to right (negative x) and bottom to top (positive y)
So the second row (above) is shifted 0.5 pixel to the left --> should be +0.5

for left-hand side modules, the first pixel is at the top right corner.
It reads right to left (positive x) and top to bottom (negative y)
So the second row (below) is shifted 0.5 pixel to the right --> should be -0.5

Does this make any sens?

takluyver: That does make sense, however...

The coordinates at this point are not x and y, but slow-scan and fast-scan directions. I.e. positive fast-scan is the first read direction (left to right for the RHS modules, right to left for the LHS). In this scheme, we want to shift the odd rows in the negative fast-scan direction (so pixel [1, 0] is moving away from [0, 1] - see the diagram above).

Further along in the code, we multiply these by the unit vectors of the grid for this module (e.g. `fs_unit_x`) to transform them into x and y coordinates. That's where they'll go in different directions for different quadrants: `fs_unit_x` will be negative for RHS modules and positive for LHS modules.

It's pretty tricky to explain this in words without a piece of paper or even waving my hands around, so let me know if that's still not clear.

tmichela: Thanks, that's actually quite clear :)

tmichela:
No further comments from my side, looks good to me

takluyver:
Thanks!

takluyver:
(Removed the TODO comment because Monica from DET confirmed by email that the layout looks right)

PR #177, opened 2019-06-19
Better error message for missing train ID

takluyver:
From https://in.xfel.eu/redmine/issues/42350

Before:

```
KeyError: 131029985
```

After:

```
TrainIDError: Train ID 131029985 not found in this data
```


fangohr:
Okay, thanks.

PR #179, opened 2019-06-19
Tiny cleanup

tmichela:
Small PR trying to polish some things before the next release, mainly to rename `DSSC` classes to `DSSC1M` to stick to the naming convention.

I also:
- removed old tests
- updated `setup.py`
- few pep8 auto-corrections and unused imports


setup.py
```diff
@@ -44,12 +44,8 @@ def find_version(*parts):
           ],
```
tmichela: in `package_data` should we also include the dssc geometry file?

takluyver:
Thanks,  this looks good. The DSSC geometry example notebook will need the name change as well.

setup.py
takluyver: Yes, good catch, thanks.

takluyver:
The test failures are because I accidentally removed `os` (overlapping pull requests - one removed it, one added code using it). If you merge from master, it should be OK.

takluyver:
Ah, seemingly you don't need to merge from master.

tmichela:
yep

karabo_data/reader.py
```diff
@@ -30,7 +30,6 @@
 from .read_machinery import (
     DETECTOR_SOURCE_RE,
     DataChunk,
-    FilenameInfo,
```
takluyver: This one was deliberately left despite being unused, in case other code outside karabo_data is using it (`from karabo_data.reader import FilenameInfo`)

tmichela: aha, I'll put it back

takluyver:
Other than that one import, LGTM

tmichela:
Thanks

PR #181, opened 2019-06-21
Restore deprecated DSSC_Geometry name for now

takluyver:
I noticed there's at least one example already circulating with the old name in use. It's easy enough to leave it there as a compatibility alias for now, and print a warning when it's used.

cc @tmichela


takluyver:
Offline LGTM from @tmichela

PR #183, opened 2019-07-04
Add --source and --key options to karabo-bridge-serve-files

takluyver:
It's more efficient to read and stream only specific data of interest rather than an entire run. This makes it possible to do that from the `karabo-bridge-serve-files` CLI.

At present, it's only possible to specify one source pattern and one key pattern to select data, so if you wanted to stream e.g. AGIPD data and an XGM together, you'd either need to write a tiny Python script to do that, or put up with streaming all sources from the CLI. If necessary, we can think about ways to specify multiple patterns, but it might be a bit awkward with command-line options.

https://in.xfel.eu/redmine/issues/42840

### DIFF
--- a/karabo_data/export.py
+++ b/karabo_data/export.py
@@ -216,7 +216,7 @@ class ZMQStreamer:
         self._buffer.put(self._serialize(data, metadata))
 
 
-def serve_files(path, port, **kwargs):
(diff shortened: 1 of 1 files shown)


tmichela:
LGTM

fangohr:
A bit late to the party: have we got examples in the documentation for this feature?

Issue #180, opened 2019-06-20
Deduplicate get_pixel_positions and to_distortion_array


takluyver:
These two methods are doing quite similar calculations. We should share some of the code between them, so there are fewer places for bugs to hide.

The differences are:

- `get_pixel_positions` returns one 3D coordinate for each pixel, but `to_distortion_array` has one for each corner.
- `to_distortion_array` shifts the origin so that no x/y coordinates are negative
- `to_distortion_array` combines the module and slow-scan dimensions, to match data passed as a 2D array.

takluyver:
- `to_distortion_array` has (z, y, x) coordinates instead of (x, y, z).

PR #184, opened 2019-07-04
Reduce duplicated code for generating distortion arrays

takluyver:
This reuses some logic which was duplicated between distortion arrays and the new `get_pixel_positions()` method (#180). The base `to_distortion_array` method is now also compatible with hexagonal pixels for DSSC, which no longer needs a separate implementation of the method.

This is slightly less code to maintain, but more importantly, fewer code paths mean there are fewer places for bugs to lurk undetected.

I reran the [DSSC geometry notebook](https://karabo-data.readthedocs.io/en/latest/dssc_geometry.html) manually to check that the pixel shapes and centres still look the same.

### DIFF
--- a/karabo_data/geometry2.py
+++ b/karabo_data/geometry2.py
@@ -418,9 +418,18 @@ class DetectorGeometryBase:
 
     @classmethod
     def _distortion_array_slice(cls, m, t):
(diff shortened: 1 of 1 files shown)


tmichela:
LGTM

Issue #165, opened 2019-06-04
Reuse output array for assembling detector image


takluyver:
Profiling the 'snapped' geometry assembly with one train (128 pulses) of LPD data, I see that nearly 40% of the total time is for allocating/nan-filling the output array (the system probably cheats and only allocates memory when it's filled). We could save this much time by passing in an `out` array of the correct shape.

This could be valuable when we are assembling detector data for many trains with the same geometry - the calling code could recycle a single big array, or swap two arrays between reading and writing.

@zhujun98 @ebadkamil - I think this scenario fits what karaboFAI does. Would it be useful for you if you could pass an array in to speed up assembling an image?

We may also be able to get some minor improvements by caching things, but most of the rest of the time is spent copying data into the output array, which I don't know how to speed up.

-----

line_profiler results:

```
Timer unit: 1e-06 s

Total time: 0.474858 s
/home/kluyvert/.local/lib/python3.6/site-packages/karabo_data/geometry2.py
Function: position_modules at line 687

Line #      Hits         Time  Per Hit   % Time  Line Contents
==============================================================
   687                                               def position_modules(self, data):
   688                                                   """Implementation for position_modules_fast
   689                                                   """
   690         1         11.0     11.0      0.0          assert data.shape[-3:] == self.geom.expected_data_shape
   691         1       5313.0   5313.0      1.1          size_yx, centre = self._get_dimensions()
   692         1        410.0    410.0      0.1          out = np.empty(data.shape[:-3] + size_yx, dtype=data.dtype)
   693         1     185049.0 185049.0     39.0          out[:] = np.nan
   694        17        149.0      8.8      0.0          for i, module in enumerate(self.modules):
   695        16        130.0      8.1      0.0              mod_data = data[..., i, :, :]
   696        16       5910.0    369.4      1.2              tiles_data = self.geom.split_tiles(mod_data)
   697       272       1139.0      4.2      0.2              for j, tile in enumerate(module):
   698       256        622.0      2.4      0.1                  tile_data = tiles_data[j]
   699                                                           # Offset by centre to make all coordinates positive
   700       256       5773.0     22.6      1.2                  y, x = tile.corner_idx + centre
   701       256       1194.0      4.7      0.3                  h, w = tile.pixel_dims
   702       256       2343.0      9.2      0.5                  xformed = tile.transform(tile_data)
   703       256     266813.0   1042.2     56.2                  out[..., y : y + h, x : x + w] = xformed
   704                                           
[...]

PR #186, opened 2019-07-09
Allow creating and reusing an output array for assembling detector images

takluyver:
As described in #165, this offers something like a 40% speedup if you're repeatedly assembling a train of images from the same detector with the same geometry. I've tested it with a train of 176 AGIPD images, and I think I'm seeing that kind of improvement, although the actual timings are rather variable.

@ebadkamil @zhujun98 - I think this is something karaboFAI could make use of. Whether or not you intend to, it would be good if you could review this code, as you're familiar with using it for this kind of scenario.

### DIFF
--- a/docs/geometry.rst
+++ b/docs/geometry.rst
@@ -50,6 +50,8 @@ but this geometry code works with a position for each tile.
 
    .. automethod:: position_modules_fast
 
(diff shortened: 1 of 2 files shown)


tmichela:
LGTM

PR #187, opened 2019-07-16
Assume .h5 files are HDF5 when opening a run

takluyver:
We already catch errors when opening the files, so we'll catch if we try to open a file that's not valid HDF5.

This speeds up opening a run by about 5% in my tests. That's a small gain, but it's low-hanging fruit - neither us nor the users need to do anything clever to use it. I'm thinking about how to get bigger speed improvements.


tmichela:
LGTM

PR #191, opened 2019-07-18
Support 2D array layouts for CrystFEL format geometry.

### LINKED ISSUE #190: Write CrystFEL geometry for 2D array
OnDA & cfelpyutils appear to only handle input data in a 2D array (per frame). The CrystFEL geometry files karabo_data writes currently describe how to assemble the data from a 3D array, with a separate axis for the 16 modules. But this can't directly be used with OnDA.

We should make it possible to write the geometry for a 2D input array, with modules arranged in a strip along the slow-scan dimension.

Discussed with @ebadkamil.

takluyver:
Fixes #190. This was easier than I expected, thanks to @antarcticrainforest's work to describe the array dimensions for `write_crystfel_geom`. To use it for a 2D array, pass:

```python
geom.write_crystfel_geom(..., dims=('frame', 'ss', 'fs')
```

The default dims has an extra entry `'modno'`. The absence of this causes it to spread modules out along the slow-scan dimension instead.

I've also moved the CrystFEL writing code to a separate module so it's easier to see. To review the real functional changes here, ignore the first two commits.

### DIFF
--- a/karabo_data/tests/test_agipd_geometry.py
+++ b/karabo_data/tests/test_agipd_geometry.py
@@ -68,6 +68,7 @@ def test_write_read_crystfel_file_2d(tmpdir):
     geom_dict = load_crystfel_geometry(path)
 
     p3a7 = geom_dict['panels']['p3a7']
+    assert p3a7['dim_structure'] == ['%', 'ss', 'fs']
     assert p3a7['min_ss'] == (3 * 512) + 448
     assert p3a7['max_ss'] == (3 * 512) + 511
     assert p3a7['min_fs'] == 0
--- a/karabo_data/tests/test_agipd_geometry.py
+++ b/karabo_data/tests/test_agipd_geometry.py
@@ -1,5 +1,6 @@
 
 from cfelpyutils.crystfel_utils import load_crystfel_geometry
+from itertools import product
 from matplotlib.axes import Axes
 import numpy as np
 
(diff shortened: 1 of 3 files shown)


takluyver:
Here's a link to view only the changes after the reorganisation:

https://github.com/European-XFEL/karabo_data/pull/191/files/bffb25a3cdb270197e751127f70eb258bd106767..57c60d9a341dd3d507b11750bda57fd8ac5dd724

karabo_data/tests/test_agipd_geometry.py
```diff
@@ -33,7 +33,7 @@ def test_write_read_crystfel_file(tmpdir):
     )
     np.testing.assert_allclose(loaded.modules[0][0].fs_vec, geom.modules[0][0].fs_vec)
 
-    # Load the geometry file with cfelpyutils and test the ridget groups
+    # Load the geometry file with cfelpyutils and test the rigid groups
     geom_dict = load_crystfel_geometry(path)
     quad_gr0 = ['p0a0', 'p0a1', 'p0a2', 'p0a3', 'p0a4', 'p0a5', 'p0a6', 'p0a7',
```
antarcticrainforest: just some synstastic sugar and I think I did this in the first place but maybe, if you want, you'll could use itertool's product to shorten this a bit ?

takluyver: Thanks, this is a good idea. Done.

antarcticrainforest:
LGTM

takluyver:
Nice, double LGTM :-D

PR #193, opened 2019-07-19
Use metres for geometry internals

### LINKED ISSUE #160: Represent detector geometry in metres instead of pixels
The `geometry2` module uses 'pixel units' for tile positions and pixel vectors, following CrystFEL's geometry format. However, this is somewhat awkward when the pixels don't have a 1:1 aspect ratio, as in DSSC, because there are different pixel units in different dimensions.

I've worked around this for DSSC by arbitrarily picking one dimension (236 µm, ⬡ hexagon width) as the pixel unit, and stretching the dimensions of the 'snapped' geometry to fit the other dimension. But it makes it harder to understand the code.

Once the DSSC geometry is working and merged, I propose to change the geometry2 code to store positions and vectors primarily in metres. This will involve a bit of extra conversion to and from CrystFEL geometry (which uses pixel units), but it better aligns with XFEL geometry (mm) and pyFAI distortion arrays (m).

I'm only planning to make this change to the precise, 3D geometry (GeometryFragment, DetectorGeometryBase and subclasses). The snapped geometry will still use pixels as the base, because it's designed for indexing into an array representing an assembled image.

takluyver:
Currently, the geometry machinery works with 'pixel units', following CrystFEL's format. This is hard to follow when pixels are not square, as for DSSC. This changes it to use metres internally.

All input and output still happens in the same units by default; this is only changing the internals. And it only changes the parts handling exact geometry: the 'snapped' geometry still works in pixels, because it's indexing into an array.

This is up for discussion. It's somewhat more code to work with, but I think it's marginally easier to think about it in metres.

Fixes #160.


tmichela:
LGTM

PR #194, opened 2019-08-02
Add log message before opening run to make virtual CXI file

takluyver:
I noticed while testing the command that there's a long pause with no hint about what it's doing as it reads the run. This will log a descriptive message before that.


takluyver:
This is a trivial change, so I'm going to merge it to include it in 0.6.

Issue #188, opened 2019-07-16
Assemble detector modules without copying them into a 'stack' array


takluyver:
From discussion relating to KaraboFAI: if you want to load data from AGIPD/LPD/DSSC and assemble it, it currently gets copied twice in memory:

individual module arrays -> stacked array -> assembled image

The first of these copies should be unnecessary. There are two basic schemes for how we could avoid it:

- Make a 'virtual' stack, e.g. a list of arrays, to represent the data without copying. We don't need to do any maths between the modules, so they don't need to be in a single array.
- Read the data directly into slices of the stacked array, so we don't allocate individual arrays for each module.

zhujun98:
I would go for the first method and make a `StackedData` class which has several methods to provide the necessary `np.array` interface, e.g. shape property:
1. The stacked data provides the similar interface as the data from the calibration pipeline, although one needs to move the axis twice for LPD and DSSC. Therefore, it is good to 'keep' it;
2. Depending on the level of backward compatibility you want to have, there is a few options to make the new `stack_detector_data` function. No matter what, you do not need to touch the method `position_modules_fast` which is used for assembling.

takluyver:
PR #196 is an initial attempt at implementing option 1.

takluyver:
Done in #196.

PR #196, opened 2019-08-02
'Virtual' stack to hold detector modules

takluyver:
To avoid an extra copy of detector data which is going to be assembled. The `VirtualStack` class is like an array with very limited functionality - but enough to pass into the image assembly code. This is meant to address #188.

Other name suggestions are welcome - 'virtual' could get confusing with HDF5 virtual datasets.

I tested this with some a train of 176 pulses from AGIPD. Stacking into a real array took ~360 ms, and assembling that took ~860 ms. Using a virtual array, stacking took << 1 ms, and assembling was roughly unchanged. So in this case it's about a 30% gain, excluding any time to load the data.


tmichela:
`StackView`?

karabo_data/reader.py
```diff
@@ -1324,18 +1313,22 @@ def stack_detector_data(train, data, axis=-3, modules=16, only='', xcept=(),
...
-    return np.moveaxis(combined, 0, axis)
+    if virtual:
+        from .read_machinery import VirtualStack
+        return VirtualStack(modno_arrays, modules, shape, dtype, fillvalue)
+    else:
+        combined = np.full((modules,) + shape, fillvalue, dtype=dtype)
+        for modno, array in modno_arrays.items():
+            combined[modno] = array
+
+        return np.moveaxis(combined, 0, axis)
```
zhujun98: I would use something like 

```py
v = VirtualStack(...)`
if virtual:
    return v
return v.asarray()
```
to avoid the duplication of code.

karabo_data/read_machinery.py
```diff
@@ -205,3 +205,80 @@ def find_proposal(propno):
...
+
+
+class VirtualStack:
+    """'Virtual' array for stacking detector data with modules on axis -3
+
+    Access is limited to either a single module at a time or all modules
+    together, but this is enough to assemble detector images.
+    """
+    def __init__(self, data, nmodules, mod_shape, dtype, fillvalue):
+        self._nmodules = nmodules
```
zhujun98: Should this have a default value which is None (`data.shape[0]`)?

karabo_data/reader.py
takluyver: Yup, that makes sense.

karabo_data/read_machinery.py
takluyver: As it's only constructed somewhere where we already have that number, I prefer to always pass it explicitly. If any of the modules are missing data, you want `nmodules > len(data)`.

zhujun98: OK! I see. Thanks!

karabo_data/reader.py
takluyver: Actually, I remember now why I didn't do this: they're logically equivalent, but different in terms of memory arrangement. The existing `stack_detector_data` code keeps the data from each module together in memory but then adjusts the axes to look like (pulses, modules, slow_scan, fast_scan). `VirtualStack.asarray()` reshuffles the data so each pulse is contiguous in memory.

I'm not sure how big a difference this makes, but I'll see if it's practical to make `VirtualStack.asarray()` do something more similar to the existing assembly.

zhujun98: I think the difference depends on the way of looping over the pixel values. Normally, one pick a pulse (in a for loop) and then loop over the pixels, then it would be nice to have all the pixels continuous in memory. Otherwise, it will need to jump to the memory of another module 16 times for one pulse and I guess the performance penalty is 10% - 20%.

Aha, here what you do is copy to the stacked data to the assembled image. Then which one is faster depends on your assembling code,

karabo_data/stacking.py
```diff
@@ -0,0 +1,208 @@
...
+    # Multidimensional slicing
+    def __getitem__(self, item):
+        if not isinstance(item, tuple):
+            item = (item,)
+
+        missing_dims = self.ndim - len(item)
+        if Ellipsis in item:
+            ix = item.index(Ellipsis)
+            missing_dims += 1
+            item = item[:ix] + (slice(None, None),) * missing_dims + item[ix+1:]
```
zhujun98: Suppose you have a 4D array x and slice it with x[(0, ..., 0)]. Then `ix` will be 1, but  I guess `item` should be

```
item = item[:ix] + (slice(None, None),) * 2 + item[ix+2:]
```
You are doing something fancy and I am trying to follow you :-)

takluyver: Slicing in Python includes the first index specified but excludes the last one (I believe this is called "half-open intervals"). So `item[:1], item[1:]` would split item with no gap and no duplicate. `item[ix+1]` leaves a gap of one element from the input, which is the `...`.

Perhaps it's clearer with an example:

```
In [1]: (0, ..., 0)[:1]                                                                 
Out[1]: (0,)

In [2]: (0, ..., 0)[2:]                                                               
Out[2]: (0,)

In [3]: (0, ..., 0)[:1] + (1, 1) + (0, ..., 0)[2:]                                      
Out[3]: (0, 1, 1, 0)
```

Does that make sense?

zhujun98: Thanks! I misled myself. `item` here is the slicing instead of the array itself.

takluyver: Yup. That was the default naming Pycharm gave when I created the `__getitem__` method, but maybe it would be clearer with `slices`.

karabo_data/stacking.py
```diff
@@ -0,0 +1,208 @@
...
+                "VirtualStack can only slice a single module or all modules"
+            )
+
+    def _get_single_mod(self, modno, mod_slices):
+        try:
+            mod_data = self._data[modno]
+        except KeyError:
+            if modno >= self._nmodules:
+                raise IndexError(modno)
+            mod_data = np.full(self._mod_shape, self._fillvalue, self.dtype)
```
zhujun98: Does it make sense also to set 
```
self._data[modno] = mod_data
```
although I don't know the use case of getting a missing module twice.

here?

takluyver: I'm not sure the use case either. It would make the behaviour more consistent if you do something like `stack[0][100:200, 100:200] = 2.0`, though. At present, the effect of that is different if module 0 is there or not.

zhujun98: This is a good reason. Thanks!

zhujun98:
LGTM! Thanks a lot for implementing this.

takluyver:
No problem. Any thoughts on the naming? I'm coming round to something like StackView or StackedView, as @tmichela suggested.

takluyver:
The class is now called `StackView`, and the parameter for `stack_detector_data()` is `real_array=False`. I'm not entirely happy with that parameter name: anyone have a better idea?

I considered `copy=False`, but I don't want to give the impression that you can assemble a real numpy array as a view of several disjoint chunks of data.

zhujun98:
Maybe `view=True`?

takluyver:
I'm not so keen on `view=True`, because Numpy already uses the term 'view' quite a lot, and it's not a numpy view we're creating.

tmichela:
I am fine with `real_array`, unless you found something you like better... LGTM

takluyver:
I haven't thought of anything better. I'll merge it with `real_array` as the parameter name. We can still change it if we have a better idea later.

PR #197, opened 2019-08-05
Add dtype for output array creation

ebadkamil:
Want to use `output_array_for_position_fast` with specific dtype which may or maynot be `np.float32` and may also change on the fly during online analysis in karaboFAI. It will be good to have dtype for this function rather than using `_snapped().make_output_array` for external tools. Or may be there is another better way.

@takluyver

### DIFF
--- a/karabo_data/geometry2/__init__.py
+++ b/karabo_data/geometry2/__init__.py
@@ -286,7 +286,7 @@ class DetectorGeometryBase:
         """
         raise NotImplementedError
 
-    def output_array_for_position_fast(self, extra_shape=()):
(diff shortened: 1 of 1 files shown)


takluyver:
Thanks; that was an oversight. I guess you want a quick release so you can use this?

ebadkamil:
I think there is no need to hurry. Probably we will not bring this funtionality in `karaboFAI 0.5.1`. We can wait till next week Monday. Jun says he may also have few more requirements for `karabo_data`.

@zhujun98

takluyver:
OK. Let me know if you do want a release sooner; I haven't made any other changes yet, so it would be easy to do 0.6.1 with just this fix.

PR #199, opened 2019-08-07
Check whether cast is safe when reusing output for position_modules

zhujun98:
@takluyver @ebadkamil 

As discussed. I cannot find the test code for this part but I think it should be fine.

### DIFF
--- a/karabo_data/geometry2/__init__.py
+++ b/karabo_data/geometry2/__init__.py
@@ -903,7 +903,9 @@ class SnappedGeometry:
             out = self.make_output_array(data.shape[:-3], data.dtype)
         else:
             assert out.shape == data.shape[:-3] + self.size_yx
-            assert out.dtype == data.dtype
+            if not np.can_cast(data.dtype, out.dtype, casting='safe'):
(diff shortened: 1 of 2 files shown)


takluyver:
I'm not sure that there's currently a test covering this particular branch. It would be good to have one, though, especially now that it allows different dtypes. See e.g. [test_agipd_geometry.py](https://github.com/European-XFEL/karabo_data/blob/master/karabo_data/tests/test_agipd_geometry.py).

I'd add a test which exercises the uint16 -> float32 conversion you want to use.

zhujun98:
I added a few lines in `test_agipd_geometry.py` to cover both safe and unsafe cast.

takluyver:
Thanks, this looks good. Do you want a quick release so you can use this feature in karaboFAI soon?

zhujun98:
A quick release of 0.6.2 is appreciated. We would like to have this enhancement in the karaboFAI release 0.5.2 next Monday.

takluyver:
OK, 0.6.2 is out now.

ebadkamil:
Thanks

PR #201, opened 2019-08-09
Rework docs for reading data

takluyver:
I've tried to make the main documentation page on reading data more approachable, breaking it up into sections with subheadings, including more explanation and more example snippets.


tmichela:
nice! LGTM

Issue #131, opened 2019-03-27
Error combining raw and proc data into one collection


takluyver:
While responding to #130, I tried to combine calibrated AGIPD data with raw data from other sources like this:

```python
from karabo_data import RunDirectory

run_proc = RunDirectory('/gpfs/exfel/exp/SPB/201701/p002012/proc/r0240')
run_raw = RunDirectory('/gpfs/exfel/exp/SPB/201701/p002012/raw/r0240')
run_raw = run_raw.deselect('*AGIPD1M*')
run = run_raw.union(run_proc)
```

The last line throws:

```
ValueError: 16 sources have conflicting data (same train ID in different files): SPB_DET_AGIPD1M-1/DET/15CH0:xtdf, ...
```

(Listing all 16 AGIPD source names)

I guess the `union()` is re-selecting the raw AGIPD data which was dropped by `deselect()`.

PR #202, opened 2019-08-20
Only include relevant files from .select() and .deselect()

takluyver:
This partly addresses #131: it should make it possible to combine proc detector data with raw data, by deselecting the detector sources in the raw data.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -810,7 +810,7 @@ class DataCollection:
         files = [f for f in self.files
                  if f.all_sources.intersection(selection.keys())]
 
-        return DataCollection(self.files, selection=selection, train_ids=self.train_ids)
+        return DataCollection(files, selection=selection, train_ids=self.train_ids)
 
     def select_trains(self, train_range):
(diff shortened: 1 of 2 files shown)


takluyver:
In the older runs, as described in #131, only data files from the fast detectors ended up in proc, so combining raw & proc was useful. In more recent experiments I just found, it looks like the `DA` files from the data aggregators are also copied to proc, so maybe this isn't so important any more. But maybe those files are being copied as a workaround to make it easier to process the data from a run together.

tmichela:
Are all `DA` files copied to proc? Because some detector data (10Hz, like Jungfrau, Epix, ...) are stored in `DA` files. So running calibration on those detector would copy also the specific `DA` file.

takluyver:
I thought about that, but it looks like files with no detectors in are also getting copied. See e.g. `/gpfs/exfel/exp/SCS/201921/p002409/proc/r0102`

tmichela:
This seems like the right thing to do anyway, regardless of what file is copied or not...
LGTM

Issue #204, opened 2019-08-24
More control of image parameters in geometry.plot_data_fast()


takluyver:
There are various options to `imshow()` which people want to use, e.g. `vmin` and `vmax` which are similar to `np.clip()`, and `norm` which allows the colour scale to be log normalised. Not being able to use these with `plot_data_fast()`, people assemble the array and `imshow()` it, which is inferior (IMO) for things like axis labels and representing NaN.

This could be resolved by accepting `**kwargs` and passing them through to `imshow()`, or an `imshow_kwargs` dictionary argument.

It would also be useful to have a colour bar alongside these plots.

antarcticrainforest:
That's a good point. Since I've just learned that the computing cluster here is on maintenance and I don't have anything better to do I'll look into that.

takluyver:
Cool, go for it. If you finish that before your cluster comes back, there are also a couple of PRs open on geoAssembler you could look at. Although maybe other people should review them to get used to that code.

takluyver:
Closed by #205.

PR #205, opened 2019-08-26
More control of image parameters in geometry.plot_data_fast()

antarcticrainforest:
This PR aims at resolving issue #204. The user can pass additional keyword arguments for an axis object (e.g subplots), controlling colorbar axis + properties and of course for controlling the image plotting itself. The ``agipd_geometry.ipynb`` notebook should demo this, hopefully - I don't have access to data.


karabo_data/geometry2/__init__.py
```diff
@@ -949,23 +973,36 @@ def plot_data(self, modules_data, axis_units='px', frontview=True):
...
 
-        ax.imshow(res, origin='lower', cmap=my_viridis, extent=extent)
+        # Use a dark grey for missing data
+        _cmap = copy(viridis)
+        _cmap.set_bad('0.25', 1.0)
+        for key, default in (('extent', _extent),
+                             ('cmap', _cmap),
+                             ('origin', 'lower')):
+            # Set some default key word arguments for the plot
+            kwargs[key] = kwargs.get(key, default)
```
takluyver: I think you can use `kwargs.setdefault()` here ([docs](https://docs.python.org/3/library/stdtypes.html#dict.setdefault)).

I'd also get rid of the for loop - it takes more lines as a loop than without one.

karabo_data/geometry2/__init__.py
```diff
@@ -359,10 +368,23 @@ def plot_data_fast(self, data, axis_units='px', frontview=True):
         frontview : bool
           If True (the default), x increases to the left, as if you were looking
           along the beam. False gives a 'looking into the beam' view.
+        ax : `~matplotlib.axes.Axes`, list of Axes, optional
+          Parent axes that will be used to draw the image.
```
takluyver: I don't think this can be a list of axes here, because we call `ax.imshow()`.

karabo_data/geometry2/__init__.py
```diff
@@ -949,23 +973,36 @@ def plot_data(self, modules_data, axis_units='px', frontview=True):
...
+            # Set some default key word arguments for the plot
+            kwargs[key] = kwargs.get(key, default)
+        extent = kwargs.get('extent', _extent)
+
+        if ax is None:
+            fig = plt.figure(figsize=figsize or (10, 10))
+            ax = fig.add_subplot(1, 1, 1)
+
+        im = ax.imshow(res, **kwargs)
+        if cbar or cbar_args or cax:
```
takluyver: It's a bit ugly that we have 3 different arguments, any of which can enable a colour bar.

Maybe we could combine them as a single `colorbar` option, taking a bool (True -> colorbar with default options), or a dict (like `cbar_args`).

takluyver: For comparison, pandas [hexbin plots](https://pandas.pydata.org/pandas-docs/stable/user_guide/visualization.html#visualization-hexbin) take a `colorbar` argument which is only True/False.

karabo_data/geometry2/__init__.py
```diff
@@ -359,10 +368,23 @@ def plot_data_fast(self, data, axis_units='px', frontview=True):
         frontview : bool
           If True (the default), x increases to the left, as if you were looking
           along the beam. False gives a 'looking into the beam' view.
+        ax : `~matplotlib.axes.Axes`, list of Axes, optional
+          Parent axes that will be used to draw the image.
+        figsize : tuple
+          Size of the figure to be drawn (default: (10, 10))
```
takluyver: Should document that it's width/height in "inches" (as defined by matplotlib)

karabo_data/geometry2/__init__.py
```diff
@@ -359,10 +368,23 @@ def plot_data_fast(self, data, axis_units='px', frontview=True):
...
+        cbar : bool
+          Whether or not to display a colorbar
+        cax : `~matplotlib.axes.Axes` object, optional
+          Axes into which the colorbar will be drawn. (default: None)
+        cbar_args : dict
+          Additional keyword arguments for customizing the colorbar
+        kwargs :
+          Additional keyword arguments for customizing the plot
```
takluyver: Document that these are passed through to matplotlib's imshow.

karabo_data/geometry2/__init__.py
takluyver: Also, it's not the *parent* axes in this context, just the axes. It's the parent for the colorbar axes.

karabo_data/geometry2/__init__.py
antarcticrainforest: What about the cax argument. Should it stay as it is?

takluyver: I'd probably drop it - you can pass `colorbar={'cax': cax}` if you need it, right?

antarcticrainforest: Ok

karabo_data/geometry2/__init__.py
antarcticrainforest: Yes, of course! That was me being lazy while copying that from the matplotlib docstring.

karabo_data/geometry2/__init__.py
antarcticrainforest: Yes

karabo_data/geometry2/__init__.py
```diff
@@ -341,7 +341,14 @@ def position_all_modules(self, data, out=None):
...
         return self.position_modules_fast(data, out=out)
 
-    def plot_data_fast(self, data, axis_units='px', frontview=True):
+    def plot_data_fast(self,
+                       data,
+                       axis_units='px',
+                       frontview=True,
+                       ax=None,
+                       figsize=None,
+                       cbar=False,
```
takluyver: I'd be open to making this `True` by default. Thoughts - @tmichela or anyone?

I'd also prefer to spell out `colorbar` - it's both clearer and consistent with the naming in matplotlib and pandas.

karabo_data/geometry2/__init__.py
```diff
@@ -359,10 +366,20 @@ def plot_data_fast(self, data, axis_units='px', frontview=True):
         frontview : bool
           If True (the default), x increases to the left, as if you were looking
           along the beam. False gives a 'looking into the beam' view.
+        ax : `~matplotlib.axes.Axes` object, optional
+          Axes that will be used to draw the image (default: None).
```
takluyver: Nitpick: the default is to create a new figure. `None` is an implementation detail.

karabo_data/geometry2/__init__.py
```diff
@@ -949,23 +967,34 @@ def plot_data(self, modules_data, axis_units='px', frontview=True):
...
+        kwargs.setdefault('cmap', _cmap)
+        kwargs.setdefault('extent', _extent)
+        kwargs.setdefault('origin', 'lower')
+
+        if ax is None:
+            fig = plt.figure(figsize=figsize or (10, 10))
+            ax = fig.add_subplot(1, 1, 1)
+
+        im = ax.imshow(res, **kwargs)
+        if cbar:
```
takluyver: This means `cbar={}` is like `cbar=False`. I think if it's a dict containing keyword arguments, I'd expect `{}` to use the defaults, i.e. like True. I don't know if that's what everyone would expect, though.

karabo_data/geometry2/__init__.py
antarcticrainforest: Ok, you're write!

karabo_data/geometry2/__init__.py
antarcticrainforest: Probably not. I was also thinking it would be nice if people would have access to the colorbar object, in case it need manipulation. I guess that would involve returning the colorbar object with the axis. But I doubt that is is a good solution. Any thoughts ?

tmichela:
@antarcticrainforest is back in the game \o/

I won't help much reviewing plotting stuffs,  I'll look at the notebook at least :)

karabo_data/geometry2/__init__.py
antarcticrainforest: Yes it does. You can get the colorbar by taking the detour through retrieving the associated image: ```im = ax.images[0]; colorbar = im.colorbar```. I've outlined that in the notebook.

takluyver:
I've refined the examples in the notebook a bit, and I'm happy with this now. @tmichela, do you still want to have a look at it as well?

tmichela:
LGTM

karabo_data/geometry2/__init__.py
```diff
@@ -341,7 +341,14 @@ def position_all_modules(self, data, out=None):
...
 
-    def plot_data_fast(self, data, axis_units='px', frontview=True):
+    def plot_data_fast(self,
+                       data,
+                       axis_units='px',
+                       frontview=True,
+                       ax=None,
+                       figsize=None,
+                       colorbar=True,
+                       **kwargs):
```
tmichela: Actually, 2 minor comments:
- style this has 1 argument per line, the `plot_data` method from the `Snapped` has several on the same line, I would align the style.
- the order of arguments are also in a different order on both method, I would try to have them in the same order unless there is a reason for that.

takluyver: Thanks! Yes, I would like the argument order to be fixed (or make the new arguments keyword-only, then it doesn't matter). I'm relaxed about coding style, but consistency never hurts.

@antarcticrainforest do you still have time to make those changes? If not, one of us will do it tomorrow.

antarcticrainforest: Yes I'll do it! But. I sorry I don't find the a different order of arguments. Can someone point me to the right spot?

antarcticrainforest: Ok I think I got it now. I think I'll make them keyword arg only

takluyver:
Thanks @antarcticrainforest !

PR #206, opened 2019-08-26
Caching file contents for runs

takluyver:
Opening a run is slow: we read a little bit of data from each file to know what sources and trains are in that file. This appears to be especially bad if the files are not already cached on that node: it seems like gpfs moves big chunks of data around. But it's slow even when the data is cached locally.

Testing with `lsxfel /gpfs/exfel/exp/SCS/201901/p002212/raw/r0070` (~400 files, 1.7 TB) , I got roughly:

- 4 minutes worst case when gpfs has to transfer data onto the node
- 21 seconds reading data from each file when they've already been cached locally
- 1.6 seconds with the cache file implemented in the first commit here
- 1.4 seconds with the cache file plus lazy opening of HDF5 files (second commit)

I was hoping that lazy opening would give a further substantial speedup. It doesn't seem to, so I might still revert it. But I want to test more if it makes a difference when files are not cached locally.

The worst case is unchanged: if the cache file isn't there it still reads metadata from each file. But hopefully that only needs to happen once. Multiple users can share the cache file - it just needs someone with write access to the proposal's scratch dir to create it.

### DIFF
--- a/karabo_data/run_files_map.py
+++ b/karabo_data/run_files_map.py
@@ -7,7 +7,7 @@ import re
 from tempfile import mkstemp
 import time
 
-from .read_machinery import DATA_ROOT_DIR
+SCRATCH_ROOT_DIR = "/gpfs/exfel/exp/"
 
 log = logging.getLogger(__name__)
 
@@ -52,15 +52,24 @@ class RunFilesMap:
 
     def map_paths_for_run(self, directory):
         paths = [osp.join(directory, 'karabo_data_map.json')]
+        # After resolving symlinks, data on Maxwell is stored in either
+        # GPFS, e.g. /gpfs/exfel/d/proc/SCS/201901/p002212  or
+        # dCache, e.g. /pnfs/xfel.eu/exfel/archive/XFEL/raw/SCS/201901/p002212
         m = re.match(
-            r'(%s/\w+/\w+/\w+)/(raw|proc)/(r\d+)/?$' % DATA_ROOT_DIR, directory
+            #     raw/proc  instr  cycle prop   run
+            r'.+/(raw|proc)/(\w+)/(\w+)/(p\d+)/(r\d+)/?$',
+            os.path.realpath(directory)
         )
         if m:
-            prop_dir, raw_proc, run_nr = m.groups()
+            raw_proc, instr, cycle, prop, run_nr = m.groups()
             fname = '%s_%s.json' % (raw_proc, run_nr)
-            paths.append(
-                osp.join(prop_dir, 'scratch', '.karabo_data_maps', fname)
+            prop_scratch = osp.join(
+                SCRATCH_ROOT_DIR, instr, cycle, prop, 'scratch'
             )
+            if osp.isdir(prop_scratch):
+                paths.append(
+                    osp.join(prop_scratch, '.karabo_data_maps', fname)
+                )
         return paths
 
     def load(self):
--- a/karabo_data/run_files_map.py
+++ b/karabo_data/run_files_map.py
@@ -126,7 +126,19 @@ class RunFilesMap:
                 log.debug("Will save cached data for %s", fname)
                 need_save = True
 
-                st = os.stat(file_access.filename)
+                # It's possible that the file we opened has been replaced by a
+                # new one before this runs. If possible, get the stat from the
+                # file descriptor, which will always be accurate. Stat-ing the
+                # filename will almost always work as a fallback.
+                try:
+                    fd = file_access.file.id.get_vfd_handle()
+                except Exception:
+                    log.warning("Can't get fd for %r, will stat name instead",
+                                fname, exc_info=True)
+                    st = os.stat(file_access.filename)
+                else:
+                    st = os.stat(fd)
+
                 self.files_data[fname] = {
                     'filename': fname,
                     'mtime': st.st_mtime,
--- a/karabo_data/run_files_map.py
+++ b/karabo_data/run_files_map.py
@@ -16,7 +16,7 @@ class RunFilesMap:
     can be stored in:
 
     - (run dir)/karabo_data_map.json
-    - (proposal dir)/scratch/karabo_data_maps/raw_r0032.json
+    - (proposal dir)/scratch/.karabo_data_maps/raw_r0032.json
     """
     cache_file = None
 
@@ -32,7 +32,7 @@ class RunFilesMap:
             prop_dir, raw_proc, run_nr = m.groups()
             fname = '%s_%s.json' % (raw_proc, run_nr)
             self.candidate_paths.append(
-                osp.join(prop_dir, 'scratch', 'karabo_data_maps', fname)
+                osp.join(prop_dir, 'scratch', '.karabo_data_maps', fname)
             )
 
         self.load()
--- a/karabo_data/run_files_map.py
+++ b/karabo_data/run_files_map.py
@@ -4,6 +4,7 @@ import os
 import os.path as osp
 import re
 from tempfile import mkstemp
+import time
 
 from .read_machinery import DATA_ROOT_DIR
 
@@ -63,6 +64,7 @@ class RunFilesMap:
         This skips over invalid cache entries(based on the file's size & mtime).
         """
         loaded_data = []
+        t0 = time.monotonic()
 
         for path in self.candidate_paths:
             try:
@@ -80,6 +82,10 @@ class RunFilesMap:
             st = os.stat(osp.join(self.directory, filename))
             if (st.st_mtime == info['mtime']) and (st.st_size == info['size']):
                 self.files_data[filename] = info
+
+        if loaded_data:
+            dt = time.monotonic() - t0
+            log.debug("Loaded cached files map in %.2g s", dt)
 
     def get(self, path):
         """Get cache entry for a file path
@@ -118,6 +124,7 @@ class RunFilesMap:
                 }
 
         if need_save:
+            t0 = time.monotonic()
             save_data = [info for (_, info) in sorted(self.files_data.items())]
             for path in self.candidate_paths:
                 try:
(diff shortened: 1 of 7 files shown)


tmichela:
I haven't checked the code yet, but do you think it would make sense to have a service scanning all new runs and writing the cache somewhere readable by everybody, like `software`?

takluyver:
Yeah, once we're happy with how it operates, it would be good to run it automatically on every new run. I'd say the ideal would be to write the map into the run directory itself, because then it doesn't matter which path you access it through (e.g. `/gpfs/exfel/exp/...` vs `/gpfs/exfel/d/raw/...`.

I'm also still thinking about different formats - e.g. can we read an HDF5 file faster than JSON?

takluyver:
HDF5 is not faster. I got load times of roughly:

- JSON: 40 ms (including converting train IDs to a numpy array, and source lists to frozensets)
- HDF5: 400 ms (also tried with the 'core' driver which reads the file into memory first - about the same as using the defaults)
- Pickle: 2 ms

Pickle also produces the smallest files (example 1.9 MB, vs 2.7 MB HDF5 or 3.8 MB JSON). But of course the files are not as easily inspectable as JSON.

For lsxfel, any performance difference is eclipsed by import time - I'm preparing another PR to improve this, but it will still be on the order of 300 ms.

tmichela:
> Yeah, once we're happy with how it operates, it would be good to run it automatically on every new run. I'd say the ideal would be to write the map into the run directory itself, because then it doesn't matter which path you access it through, e.g. /gpfs/exfel/exp/... vs /gpfs/exfel/d/raw/....

Yes, that would be best to have it in the run directory itself, but this has additional complications... :)
But if we can do this in the end, maybe it makes more sens to have it in HDF5 format, even if it is not the best format for this. I also remember that IT wanted to have such a file containing only run metadata, do you have any new on this?

takluyver:
I emailed them a couple of weeks ago to ask about that, and suggest that if we write the code they could set it up to run on each run. Haven't heard anything back.

I think the performance difference is big enough that I'd prefer to avoid HDF5 for this purpose unless someone makes it a hard requirement. Unless you can see a more efficient way to use it - I've put the code I've been playing with [in a gist](https://gist.github.com/takluyver/93f05a6c7ec45fe51a4c2dd474a2dbb9), and there are sample files in `/gpfs/exfel/exp/SCS/201901/p002212/scratch/.karabo_data_maps`.

karabo_data/run_files_map.py
```diff
@@ -0,0 +1,108 @@
...
+log = logging.getLogger(__name__)
+
+class RunFilesMap:
+    """Cached data about HDF5 files in a run directory
+
+    Stores the train IDs and source names in each file, along with some
+    metadata to check that the cache is still valid. The cached information
+    can be stored in:
+
+    - (run dir)/karabo_data_map.json
```
tmichela: For `raw` directory, no-one has write access. Do you know how it is for the `proc`? Do users can write to this directory or is it only for facility services?

karabo_data/run_files_map.py
```diff
@@ -0,0 +1,108 @@
...
+        if (dirname == self.directory) and (fname in self.files_data):
+            return self.files_data[fname]
+
+        return None
+
+    def save(self, files):
+        """Save the cache if needed
+
+        This skips writing the cache out if all the data files already have
+        valid cache entries. It is also silences permission errors from writing
```
tmichela: `It also silences ...`

karabo_data/run_files_map.py
takluyver: Users can't write to `proc` as far as I know.

I've included this location in the hope that if we've got the code, we can persuade ITDM to create the map file automatically when each run is written. Then no-one needs to wait several minutes to see what's in a run.

tmichela: OK, that makes sens :+1:

fangohr: I am optimistic this could be run as a service at the end of recording a run (eventually).

[We should also push for karabo-data-validate to run automatically.]

I see the attraction of JSON files: faster reading and plain text. I also see the attraction of sticking to one file format. This is a tough call.

To run lsxfel, is this Karabo_data_map the only file that needs to be read? If so, then I guess you can 'feel' the difference between a 400ms and a 40ms read time?

takluyver: If you `lsxfel` a run directory with detector data (AGIPD/LPD/DSSC), at present it will still open a single HDF5 file from that run to get some statistics like module size and frames per train. I'm thinking of adding this into the map file too, so it only needs to read that.

With this PR plus #207, I got timings down to about 0.3s for `lsxfel` on a large run. Using HDF5 for the cache would presumably make that about 0.6s. There's a rule of thumb that things happening in 0.1s feel instant to the user, so the difference would be noticeable.

There are probably ways to make an HDF5 map file that can be read faster - combine data into larger datasets so we can do fewer, bigger reads. But I've not come up with any way to do that and preserve convenient inspectability.

tmichela:
LGTM

takluyver:
Not merging this yet, because:

1. I'm still considering putting the detector info into the cache: on the plus side, it saves opening and reading data from even one HDF5 file, but it does introduce an extra level of complexity into the cache for a specific purpose (lsxfel, not every time a run is opened).
2. I want to ensure @fangohr has time to properly consider different format options before we merge this.

takluyver:
From discussion with @fangohr this morning:

- [ ] Extend karabo-data-validate to check files map
- [ ] Some way to show timing information while using karabo_data (logging?)
- [ ] Document location & contents of map files in case they go wrong.

karabo_data/validation.py
```diff
@@ -160,6 +162,22 @@ def check_files(self):
             fv = FileValidator(f)
             self.problems.extend(fv.run_checks())
 
+    def check_files_map(self):
```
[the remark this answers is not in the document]
takluyver: Good question. My thinking is that if there's anything weird about accessing data, `karabo-data-validate` will try to find possible causes - including incorrect cache files. Also, if we can get the cache created automatically when the files are migrated/calibrated, it effectively becomes part of the data format.

Maybe we can have a single tool, but with an extra option, e.g. `--check-cache`, to also include checking the cache file.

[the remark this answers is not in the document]
tmichela: I see, it's probably correct to check them as part of the validation then, fine with me to leave it as it is.

Can't you get the file descriptor with: `file.id.get_vfd_handle()`?

Regarding the cache files 'editing', maybe we could change the file mode and set them as read-only by default, I believe that `karabo_data` will never have to edit them either, right?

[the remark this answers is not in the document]
takluyver: Oh nice, I had missed that. If that works, we should be good - HDF5 locks files while it writes them, so if we can open a file it shouldn't be changing underneath us, and if we use the fd it can't be replaced before we `stat` it.

takluyver:
It's working with getting an fd from HDF5.

However, while testing it I ran into a limitation that I thought wouldn't really matter. If you make a symlink to access data directories more conveniently, it doesn't recognise that it's a standard data directory, so it doesn't use the cache. I have two ideas at the moment to solve this:

1. Assume any path which ends like `/raw/r0123` or `/proc/r1234` is part of a proposal directory, and use `(prop_dir)/scratch/.karabo_data_maps` relative to those.
2. Resolve any symlinks with `os.path.realpath()` and check against the real data locations on Maxwell. This is unfortunate because the real paths change (from `/gpfs/exfel/d/...` to `/pnfs/xfel.eu/...`), and they're meant to be an internal implementation detail.

karabo_data/validation.py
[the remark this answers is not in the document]
RobertRosca: Seems a bit like overkill, but if people manually editing the cache file is a concern we could save a hash of the file along with it and check that the file matches the hash before loading it. I agree it's not necessary though, don't think anybody would try to edit them.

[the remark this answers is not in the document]
takluyver: That's true. I keep thinking about this like a security question, i.e. if someone can edit the file, they can also regenerate the hash. But of course we're thinking about unintentionally breaking things.

In any case, I think it's pretty unlikely that people would edit these files. And if we can get them automatically generated in the run folders, users won't have access to modify them anyway.

tmichela:
If we resolve the simlink, the path should always be `.../(raw|proc)/INSTRUMENT/CYCLE/PROPOSAL/RUN`, right?

If that's the case, solution `2.` has my favor. We don't really care if the location change as long as the end of the resolved path does not change.

tmichela:
e.g.

`/pnfs/xfel.eu/exfel/archive/XFEL/proc/SPB/201701/p002012/r0001`

`/gpfs/exfel/d/raw/SPB/201901/p002316/r0001`

takluyver:
That's an improvement,  but I think it still requires hardcoding an absolute path to find the scratch directory where we store the cache - because scratch stays on gpfs even when the data moves to `/pnfs`.

takluyver:
This seems to be working.

The p002212 raw files we were testing with have been migrated to dCache today. It seems like moving data to dCache changes the file mtimes, so the cache has to be regenerated after that. That's unfortunate, but we can live with it.

If we can get the cache file generated whenever a run is written, we'll have to ensure it's regenerated when the data is moved to dcache, otherwise the cache file will get copied across with the wrong mtimes and then we'll never use it. :-(

tmichela:
> If we can get the cache file generated whenever a run is written, we'll have to ensure it's regenerated when the data is moved to dcache, otherwise the cache file will get copied across with the wrong mtimes and then we'll never use it. :-(

That's something we should clarify with IT, but if the cache files are part of the run directory, then it cannot be modified or deleted and there is no need for checking `mtime` anymore, right? (this only applies for raw data of course)

takluyver:
In principle, yes. I'm wary of writing any cache implementation that assumes the data files could never chnage without the cache being invalidated. But maybe we can make a pragmatic compromise.

tmichela:
LGTM!

PR #209, opened 2019-09-09
Avoid memory copy in get_array

takluyver:
`get_array()` previously worked by getting a numpy array for each chunk of data (e.g. each file) and then concatenating them together, which involves a memory copy. This change makes it work out the shape first, create a numpy array of the desired size, and read the data directly into that.

Tests with ~3 MB of XGM data spread across 7 files show about a 50% speed-up with a warm cache (48 ms -> 24 ms). Presumably, if the data is cached in OS memory, this reflects that it's being copied once rather than twice.

Obviously the speed up will be much smaller in most real use cases when you're getting data that's not already cached. But it could also reduce memory pressure with large amounts of data - the previous implementation would briefly need twice as much memory.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -622,11 +622,15 @@ class DataCollection:
             if not isinstance(roi, tuple):
                 roi = (roi,)
 
-        chunks = sorted(self._find_data_chunks(source, key), key=lambda x: x.train_ids[0])
+        chunks = sorted(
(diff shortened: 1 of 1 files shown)


takluyver:
Extra benefit: if you try to load data that won't fit in memory, it should fail quickly as it tries to allocate the array, rather than loading part of it.

tmichela:
I forgot to comment before my leave. LGTM!

PR #210, opened 2019-09-09
Permission error

tmichela:
Opening a run with `RunDirectory` or `open_run` where the user does not have read access would result in a "No HDF5 files found" exception and I find that misleading.

It would now raise a `PermissionError` in such case.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -1205,12 +1205,12 @@ def RunDirectory(path):
     path: str
         Path to the run directory containing HDF5 files.
     """
-    if not os.access(path, os.R_OK):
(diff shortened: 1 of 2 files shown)


takluyver:
Thanks! I thought I remembered fixing this by using `os.listdir()` instead of glob, but apparently my memory is faulty.

Is it worth using `os.listdir()` and relying on the error produced by the system rather than throwing the error ourselves? I'm never entirely sure if `os.access` does the right thing with the GPFS access control system.

tmichela:
Yes, that's better. I use `listdir` now.

takluyver:
Thanks!

PR #212, opened 2019-09-13
Dask integration

takluyver:
[Dask](https://docs.dask.org/en/latest/) seems like exactly what we want in many ways. It can represent data like a numpy array without loading it all into memory, and then coordinate parallel computations working with chunks of the data, so you can do `arr.mean()` and it will organise tasks to load each chunk, sum it, add up the results, and give you a mean. A plugin [dask_jobqueue](https://jobqueue.dask.org/en/latest/) even makes it easy to spread across HPC clusters.

Unfortunately, my initial tests with it were not as good as I had hoped ([experiments repo](https://git.xfel.eu/gitlab/dataAnalysis/data-reading-performance)). The default settings (using multiple threads) were slower than doing the averaging without parallelism. Using dask.distributed with multiple processes on one node was better, but strangely variable, and produced a bunch of concerning warnings and errors. It wasn't maxing CPUs like other approaches I tried.

I haven't had time to dig into these issues in detail. I know dask has tools for visualising task graphs and computational progress, which hopefully can help shed some light. Two possible sources of issues:

- h5py & HDF5 have global per-process locking (like the GIL), so they can't readily be accelerated by multiple threads - and lock contention could make it slower.
- I'm using dask arrays wrapped in xarrays (for labelled axes). This [is supported](https://xarray.pydata.org/en/stable/dask.html), but it's an extra layer of complexity.

[API question: `get_dask_array(...)`? `get_array(..., dask=True)`? `get_array(..., format='dask')`?]

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -664,7 +664,6 @@ class DataCollection:
 
     def get_dask_array(self, source, key, extra_dims=None):
         import dask.array as da
-        import xarray
         chunks = sorted(
             self._find_data_chunks(source, key),
             key=lambda x: x.train_ids[0] if x.train_ids.size else 0,
@@ -690,18 +689,7 @@ class DataCollection:
                 da.from_array(chunk.dataset, chunks=chunk_shape)[chunk.slice]
             )
 
-        darr = da.concatenate(chunks_darrs, axis=0)
-
-        # Dimension labels
-        if extra_dims is None:
-            extra_dims = ['dim_%d' % i for i in range(darr.ndim - 1)]
-        dims = ['trainId'] + extra_dims
-
-        coords = {}
-        if darr.shape[0]:
-            coords = {'trainId': np.concatenate(chunks_trainids)}
-
-        return xarray.DataArray(darr, dims=dims, coords=coords)
+        return da.concatenate(chunks_darrs, axis=0)
 
     def union(self, *others):
         """Join the data in this collection with one or more others.
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -662,7 +662,27 @@ class DataCollection:
             sorted(non_empty, key=lambda a: a.coords['trainId'][0]), dim='trainId'
         )
 
-    def get_dask_array(self, source, key, extra_dims=None):
+    def get_dask_array(self, source, key):
+        """Get a Dask array for the specified data field.
+
+        Dask is a system for lazy parallel computation. This method doesn't
+        actually load the data, but gives you an array-like object which you
+        can operate on. Dask loads the data and calculates results when you ask
+        it to, e.g. by calling a ``.compute()`` method.
+        See the Dask documentation for more details.
+
+        If your computation depends on reading lots of data, consider creating
+        a dask.distributed.Client before calling this.
+        If you don't do this, Dask uses threads by default, which is not
+        efficient for reading HDF5 files.
+
+        Parameters
+        ----------
+        source: str
+            Source name, e.g. "SPB_DET_AGIPD1M-1/DET/7CH0:xtdf"
+        key: str
+            Key of parameter within that device, e.g. "image.data".
+        """
         import dask.array as da
         chunks = sorted(
             self._find_data_chunks(source, key),
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -687,7 +687,7 @@ class DataCollection:
                 chunk_shape = (chunk_dim0,) + chunk.dataset.shape[1:]
 
             chunks_darrs.append(
-                da.from_array(chunk.dataset, chunks=chunk_shape)
+                da.from_array(chunk.dataset, chunks=chunk_shape)[chunk.slice]
             )
 
         darr = da.concatenate(chunks_darrs, axis=0)
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -689,12 +689,8 @@ class DataCollection:
             key=lambda x: x.train_ids[0] if x.train_ids.size else 0,
         )
 
-        chunks_trainids = []
         chunks_darrs = []
         for chunk in chunks:
-            chunks_trainids.append(
-                self._expand_trainids(chunk.counts, chunk.train_ids)
-            )
             chunk_dim0 = np.sum(chunk.counts)
             chunk_shape = (chunk_dim0,) + chunk.dataset.shape[1:]
 
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -662,6 +662,47 @@ class DataCollection:
             sorted(non_empty, key=lambda a: a.coords['trainId'][0]), dim='trainId'
         )
 
+    def get_dask_array(self, source, key, extra_dims=None):
+        import dask.array as da
+        import xarray
+        chunks = sorted(
+            self._find_data_chunks(source, key),
+            key=lambda x: x.train_ids[0] if x.train_ids.size else 0,
+        )
+
+        chunks_trainids = []
+        chunks_darrs = []
+        for chunk in chunks:
+            chunks_trainids.append(
+                self._expand_trainids(chunk.counts, chunk.train_ids)
+            )
+            chunk_dim0 = np.sum(chunk.counts)
+            chunk_shape = (chunk_dim0,) + chunk.dataset.shape[1:]
+
+            # Find chunk size of maximum 1 GB
+            # TODO: optional user control of chunking
+            limit = 1024 ** 3
+            while np.product(chunk_shape) > limit and chunk_dim0 > 1:
+                chunk_dim0 //= 2
+                chunk_shape = (chunk_dim0,) + chunk.dataset.shape[1:]
+
+            chunks_darrs.append(
+                da.from_array(chunk.dataset, chunks=chunk_shape)
+            )
+
+        darr = da.concatenate(chunks_darrs, axis=0)
+
+        # Dimension labels
+        if extra_dims is None:
+            extra_dims = ['dim_%d' % i for i in range(darr.ndim - 1)]
+        dims = ['trainId'] + extra_dims
+
+        coords = {}
+        if darr.shape[0]:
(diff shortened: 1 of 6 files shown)


karabo_data/reader.py
```diff
@@ -662,6 +662,47 @@ def get_array(self, source, key, extra_dims=None, roi=by_index[...]):
...
+            limit = 1024 ** 3
+            while np.product(chunk_shape) > limit and chunk_dim0 > 1:
+                chunk_dim0 //= 2
+                chunk_shape = (chunk_dim0,) + chunk.dataset.shape[1:]
+
+            chunks_darrs.append(
+                da.from_array(chunk.dataset, chunks=chunk_shape)[chunk.slice]
+            )
+
+        darr = da.concatenate(chunks_darrs, axis=0)
```
antarcticrainforest: I am by no means an expert in dask as I for myself have just recently started using it, but couldn't you instantiate a dask.distributed client (with optional user given settings) when creating RunDir and then use it here?
I mean something like this:
```
with dask.config.set(self.dask_client.get):
       chunks_darrs.append(
                da.from_array(chunk.dataset, chunks=chunk_shape)[chunk.slice]
            )
        darr = da.concatenate(chunks_darrs, axis=0)
```
or 
``` darr = self.dask_config.sbmit(da.concatenate, chunks_arss, axis=0).result() ```

Maybe you only need to create the dask client once and all dask calls will be redirected to the client automatically, I am not sure? Regardless I think it would make sense to create in instance of a dask client which can be configured by the user (e.g. reserved computing node(s)).

Comment (takluyver) [reply karabo_data/reader.py]: As creating a client automatically configures Dask to use it, I figured it was best not to do that in library code, but instead recommend that user code does it. This is how I've been experimenting with it:

```python
client = Client(...)
run.get_dask_array(...)
```

I thought this would allow expert users to use dask however they prefer, rather than karabo_data forcing a particular use pattern. But I'm also new to dask, so I don't know if there's a better design.

One particular doubt: with no client set up, dask defaults to using multithreading. But HDF5 & h5py don't parallelise in threads, so I think data reading is still effectively serialised. Maybe we should do something to avoid or warn about that. But the impact depends on the relative amounts of I/O and compute you want to do.

karabo_data/reader.py
```diff
@@ -662,6 +662,51 @@ def get_array(self, source, key, extra_dims=None, roi=by_index[...]):
...
+        """Get a Dask array for the specified data field.
+
+        Dask is a system for lazy parallel computation. This method doesn't
+        actually load the data, but gives you an array-like object which you
+        can operate on. Dask loads the data and calculates results when you ask
+        it to, e.g. by calling a ``.compute()`` method.
+        See the Dask documentation for more details.
+
+        If your computation depends on reading lots of data, consider creating
+        a dask.distributed.Client before calling this.
```
antarcticrainforest: What about adding a recipe on how to use the client or even dask-jobqueue  for slurm integration to the docs?

karabo_data/reader.py
```diff
@@ -662,6 +662,51 @@ def get_array(self, source, key, extra_dims=None, roi=by_index[...]):
...
+
+        chunks_darrs = []
+        for chunk in chunks:
+            chunk_dim0 = np.sum(chunk.counts)
+            chunk_shape = (chunk_dim0,) + chunk.dataset.shape[1:]
+
+            # Find chunk size of maximum 1 GB
+            # TODO: optional user control of chunking
+            limit = 1024 ** 3
+            while np.product(chunk_shape) > limit and chunk_dim0 > 1:
```
antarcticrainforest: This is maybe just me being a slow thinker but you could add a comment here telling what you're trying to achieve.

karabo_data/reader.py
```diff
@@ -662,6 +662,51 @@ def get_array(self, source, key, extra_dims=None, roi=by_index[...]):
             sorted(non_empty, key=lambda a: a.coords['trainId'][0]), dim='trainId'
         )
 
+    def get_dask_array(self, source, key):
```
antarcticrainforest: What about making this the default method for get_array ?

karabo_data/tests/test_reader_mockdata.py
```diff
@@ -421,6 +421,18 @@ def test_run_get_virtual_dataset_filename(mock_fxe_raw_run, tmpdir):
...
+def test_run_get_dask_array(mock_fxe_raw_run):
+    import dask.array as da
+    run = RunDirectory(mock_fxe_raw_run)
+    arr = run.get_dask_array(
+        'SA1_XTD2_XGM/DOOCS/MAIN:output', 'data.intensityTD',
+    )
+
+    assert isinstance(arr, da.Array)
+    assert arr.shape == (480, 1000)
+    assert arr.dtype == np.float32
```
antarcticrainforest: Can you also check of the chunk size here?

setup.py
```diff
@@ -65,6 +65,7 @@ def find_version(*parts):
               'sphinxcontrib_github_alt',
           ],
           'test': [
+              'dask[array]',
```
antarcticrainforest: Just out of curiosity: what does that mean/do?

takluyver:
I think this is ready for review now.

My sample use case (averaging a lot of DSSC data) can still be done slightly faster with custom multiprocessing code than using a dask array. I think this is partly because with dask, the parent process has to open the files to work out the array shape, whereas you can avoid this by writing the parallel code by hand. But they're on a similar time scale, and using dask is very convenient - it's basically like using numpy plus some boilerplate.

karabo_data/reader.py
takluyver: Yes, good point. I can turn the script I've been testing with into an example.

karabo_data/reader.py
takluyver: Is "Find chunk size of maximum 1 GB" not it?

In fact, that's not exactly what it's doing, because I don't take into account the size of each entry, so it's 1 billion array elements. I should probably fix that.

takluyver:
Also, I meant to explain: for now this returns a 'plain' dask array, without xarray labels. Xarray [does have some support for dask](https://xarray.pydata.org/en/stable/dask.html), and I initially tried to return a labelled array, like `.get_array()`. But it seemed to be slower, and dask produced warnings about garbage collection taking a long time. I don't know if I was doing something wrong, or hitting a bug in one of those libraries, but ultimately I backed off the integration for now.

setup.py
takluyver: Dask defines a number of 'extras', i.e. optional sets of dependencies. So just installing `dask` gets you the core library, and omits some other packages it needs for array support.

I've put it in as a test requirement, rather than a full requirement of karabo_data, because I'm wary of loading it up with too many requirements, and it seems reasonable that users who explicitly ask for a dask array will ensure dask is installed.

karabo_data/reader.py
antarcticrainforest: Yes that's exactly what I had in mind. You could make  'xarray' the default and then print a future warning, if 'xarray' was set, that dask will become default. Just an idea though

setup.py
antarcticrainforest: Interesting, I did not know that

karabo_data/reader.py
takluyver: I think I'd leave getting a real array as the default - it's more straightforward. But maybe it could have a warning to suggest using dask if you load more than X amount of data, where X is probably some 10s of GB.

karabo_data/tests/test_reader_mockdata.py
takluyver: Probably not conveniently, because a dask array can have chunks of differing sizes, e.g. if the last sequence file doesn't have enough data for a full-size chunk.

takluyver:
I've added an example without the output for now. It was working as I was writing it, but when I came to run through it in order, the cluster was too busy for me to get a node.

tmichela:
I tried to run the notebook on maxwell. it works but I could not open the dashboard page, I get: `
500 : Internal Server Error`.

Is there something I'm doing wrong?:

tmichela:
Do you think it would be useful to also add a method for a `get_dask_dataframe`?

tmichela:
I found this article from one of the authors of Dask: https://matthewrocklin.com/blog/work/2017/09/18/pangeo-1

There are some insights of the limitations in using Xarray with dask and benchmarking of dask on a HPC. I had a quick look in the documentation but is not entirely clear to me how much data is transferred between processes and if this can be a bottleneck...

antarcticrainforest:
> I tried to run the notebook on maxwell. it works but I could not open the dashboard page, I get: ` 500 : Internal Server Error`.
>   
> Is there something I'm doing wrong?:

Did you ssh into the cluster and forwarded the port where the dask client is running?

takluyver:
I didn't need any

takluyver:
I didn't need any port forwarding, but it looks like I've installed dask under my home directory, so maybe it's something that works with a newer version. I have dask 2.3.0 and distributed 2.3.2.

My initial thinking is that dask dataframes aren't likely to be so useful - where our data fits into a columnar dataframe format, it's small enough that we don't need Dask. But maybe we'll find a use for it.

I'll have a read of Matt Rocklin's post.

tmichela:
I use the default python from max-jhub:
```
dask==2.4.0
dask-jobqueue==0.6.3
dask-labextension==1.0.3
distributed==2.4.0
```

takluyver:
Data transfers between machines will depend on what you're doing (of course). I've been experimenting with averaging data across trains, which should be relatively efficient if Dask works as I think it should.

It looks like the work that blog post discussed about HPC systems turned into dask_jobqueue, which we're using in the example.

karabo_data/reader.py
takluyver: Basically arbitrary. I did briefly play with making the chunks bigger and smaller ([performance testing log](https://git.xfel.eu/gitlab/dataAnalysis/data-reading-performance/blob/master/results.txt)), and maybe bigger was slightly better, but the timings vary so much that it's really hard to evaluate changes that don't make a massive difference.

tmichela:
LGTM

PR #216, opened 2019-10-09
Start writing release notes for 0.7

takluyver:


docs/changelog.rst
```diff
@@ -1,6 +1,37 @@
...
+  correctly raises PermissionError (:ghpull:`210`).
+- :func:`~.stack_detector_data` has a new parameter ``real_array``. Passing
+  ``real_array=False`` avoids copying the data into a temporary array on the way
+  to assembling images with detector geometry (:ghpull:`196`).
+- Importing ``karabo_data`` is faster, as packages like xarray and pandas are
+  now only loaded if you use the relevant methods (:ghpull:`207`).
+- :meth:`~.DataCollection.get_array` is slightly faster, as it avoids copying
+  data in memory unnecessarily (:ghpull:`209`)
+- When you select sources with :meth:`~.DataCollection.select` or
+  :meth:`~.DataCollection.select`, the resulting DataCollection no longer
```
tmichela: you wrote `DataCollection.select` twice --> `deselect`.

docs/changelog.rst
```diff
@@ -1,6 +1,37 @@
...
+---
+
+Data access
+~~~~~~~~~~~
+
+- Trying to open a run directory to which you don't have read access now
+  correctly raises PermissionError (:ghpull:`210`).
+- :func:`~.stack_detector_data` has a new parameter ``real_array``. Passing
+  ``real_array=False`` avoids copying the data into a temporary array on the way
+  to assembling images with detector geometry (:ghpull:`196`).
```
fangohr: Is this explained further in the documentation? Should we point to the documentation if so? (I didn't find it clear just from the statement here.)

fangohr:
Thanks for adding change notes - this is really useful.

docs/changelog.rst
takluyver: A bit - the `:func:` reference will link to the [function documentation](https://karabo-data.readthedocs.io/en/latest/agipd_lpd_data.html#karabo_data.stack_detector_data).

It's a tricky thing to explain clearly, but the key points are:

- Passing `real_array=False` is faster if it works
- It might not work everywhere that `real_array=True` does (so True is the default)
- But it should work for stacking detector data.

PR #219, opened 2019-10-21
Only inspect one file for detector_info

takluyver:
This approximates the total number of frames in the run - but that wasn't the most useful statistic anyway, because it can vary among detector modules, and we were picking an arbitrary module.

In combination with #206, this should mean we only open one HDF5 file for lsxfel (once the run map is cached), rather than all files for that detector module. Long runs can have 20 sequence files, so this should be much faster.

### DIFF
--- a/karabo_data/reader.py
+++ b/karabo_data/reader.py
@@ -1051,7 +1051,7 @@ class DataCollection:
             dims = ' x '.join(str(d) for d in dinfo['dims'])
             print("  e.g. module {} : {} pixels".format(module, dims))
             print("  {}".format(mod_source))
-            print("  {} frames per train, ~{} total frames".format(
+            print("  {} frames per train, up to {} frames total".format(
                 dinfo['frames_per_train'], dinfo['total_frames']
             ))
         print()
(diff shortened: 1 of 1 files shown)


takluyver:
At some point we'll need to be smarter if reduced data has varying numbers of frames saved for different trains, but I'm not trying to tackle that here.

karabo_data/reader.py
```diff
@@ -1049,7 +1050,8 @@ def info(self):
             module = ' '.join(mod_key)
             dims = ' x '.join(str(d) for d in dinfo['dims'])
             print("  e.g. module {} : {} pixels".format(module, dims))
-            print("  {} frames per train, {} total frames".format(
+            print("  {}".format(mod_source))
+            print("  {} frames per train, ~{} total frames".format(
```
tmichela: maybe `up to ### frames in total`?

tmichela:
Minor comment, LGTM

karabo_data/reader.py
takluyver: Thanks, I've made that change.

PR #225, opened 2020-03-09
deprecation warning

tmichela:


tmichela:
@takluyver

karabo_data/__init__.py
```diff
@@ -41,6 +41,11 @@
 from .stacking import *
 from .utils import *
 
+
+warn("The karabo-data package is deprecated, please use extra-data and extra-geom instead",
```
takluyver, line 45:
```suggestion
warn("The karabo_data package is deprecated, please use extra_data and extra_geom instead",
```

May as well use the importable names, making it easier to copy & paste.

takluyver:
Other than that, LGTM

